PR #20, opened 2020-05-18
Create core package

destogl:
This is the first version which compiles... 

Now I will start to add functionalities and "fill" the classes.


destogl:
The code is splited in #21 and in ros-controls/ros2_control#80

PR #21, opened 2020-05-20
First description of `DemoRobot` and test node for it.

destogl:
This is cleaned up version from #20. Now there is only `DemoRobot` code.


destogl:
Hi guys,

since our last meeting I worked on the implementation of the proposed concept. Here are some ideas for Demo Robot in a "standard" and minimal version. 

The minimal version should be very similar to YAML description of the robot for ROS1, except that a user can define interface type per joint.

What do you think? @bmagyar @olivier-stasse @Karsten1987

olivier-stasse:
Dear @destogl, thanks for the starting point. Could you reedit your commits through a rebase interactive such we just see the final result of your modifications and tests ?
Right now it is difficult to make relevant points when there is a whole history of changes through various commits.

destogl:
Hi @olivier-stasse, I will complete this by the end of the week

destogl:
Hi @olivier-stasse ,

i just added new PR #35, where I tried to clean everything and leave only minimal robot. Until the next meeting, I will focus on this following the proposal from @Karsten1987 in ros-controls/ros2_control#80. This weekend I am planning to prepare a "roadmap" to realize this without many braking changes as in the first proposal. Tomorrow I have a bit more time to work on this and I hope than to have something for you to review.

Have a nice weekend!

Issue #14, opened 2020-05-06
Concept for loading the Robot using `ResourceManager`


destogl:
Finalize the demo_resource_manger to support all scenarios from the [roadmap example](https://github.com/destogl/roadmap/blob/control_components_design/design_drafts/components_architecture_and_urdf_examples.md) - see ros-controls/roadmap#17

Test the following:
- [x] Loading of all robots
- [x] Simple loop with read and write calls

destogl:
the initial version is in #21 file: ros2_control_demo_robot_manager.cpp

destogl:
the concept is proposed in this [file](https://github.com/ros-controls/ros2_control_demos/blob/89dbaefca4a82a6df3d9d9352d1c2d6a17f5405a/ros2_control_demo_robot/src/ros2_control_manager_node.cpp) and it will be moved to `controller_manager` package

Issue #5, opened 2020-05-06
Create description for `demo_robot`


destogl:
* The descriptions follow the proposal from [roadmap design document](https://github.com/destogl/roadmap/blob/ros2_hardware_interface_design_proposal/design_drafts/hardware_interface.md)

Do the following:
* [ ] Define the `demo_robot` and describe it in short in README.md
* [ ] Create simple URDF to vizualize the `demo_robot`
* [ ] Create description for `ros2_control`-`Robot` class which uses "batch"-`headless` interface toward all its joints (i.e., `DemoRobotHeadless` class).
* [ ] Create description for `ros2_control`-`Robot` class which uses modular-`headless` interface toward its joints (i.e., `DemoSensorHeadless` and `DemoActuatorHeadless` classes).

destogl:
Check descriptions of
demo_robot.yaml
demo_robot_minimal.yaml
 in #21
@olivier-stasse : how does this look to you?

destogl:
@olivier-stasse here is something I would need help with, so if you have some time I would be very thankful :)

### Current state
* in #4 there is initial implementation of DemoRobot with URDF and xacro file
* I had problems with parsing xacro files in ROS2, therefore is comment about that in ros2_control_demo_robot/launch/minimal_robot.launch.py 
* I solved it by generating URDF in ROS1 and then using it directly

### What are next steps
- [ ] Use xacro file directly in ROS2
- [ ] Launch also robot_state_publisher
- [ ] Visualizes the `MinimalRobot` in rviz and check if the description make sense (test using joint_state_publisher is probably the easiest)
- [ ] Provide some fake funcitonality in `DemoRobotHardwareMinimal` class

When this works we should than concentrate on manager node and adding another examples.

Is this OK for you?

olivier-stasse:
Sounds good !
I will give it try.

Karsten1987:
I don't think it makes sense to parse xacro files. I think it makes sense to expect the `robot_description` to be presented as xml and therefore be parsed as such.

destogl:
Oh, I wrote that confusing...

I meant not to parse xacro file manually, but to convert it automatically in launch file into URDF file.

this didn‘t work for me with xacro in ROS2 for unknown reasons...

destogl:
Full list of examples: [source](https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md)

[...]

PR #35, opened 2020-06-12
Demonstration of hardware and components use in ResourceManager

destogl:
Here should be the first implementation of demo robots

The tasks are following:
- [ ] Create descriptions (#5)
- [ ] Add demo `ResourceManager` (#14)
- [ ] Add all the missing Component types +(ros-controls/ros2_control#131 and ros-controls/ros2_control#123)
- [ ] Create implementation of hardware defined in the [roadmap example](https://github.com/destogl/roadmap/blob/control_components_design/design_drafts/components_architecture_and_urdf_examples.md) - see ros-controls/roadmap#17
- [ ] Simple execution with read and writes calls to components and the hardware


destogl:
After ros-controls/ros2_control#121 and ros-controls/ros2_control#127 there are many conceptual changes. Therefore, closing this in favor of a new clean PR

PR #36, opened 2020-09-10
Remove packages intialy created, but not needed anymore

destogl:


destogl:
Coverage is failing because some packages are removed... merging...

bmagyar:
:+1:

PR #39, opened 2020-09-21
Fix compile errors

AndyZe:
A very minor PR to fix some build errors.


destogl:
I don't get this error... Thanks for PR, I will leave ti open until we do not confirm one or another way...

destogl:
This is related to #41. So it is solved by ros-controls/ros2_control#147     [check here](https://github.com/ros-controls/ros2_control/pull/147/files#diff-04c36822f25b0598bc347395128f5d86R65)

PR #58, opened 2021-02-01
Reduce wordiness, add hints for those coming from ros(1)_control

AndyZe:
- Remove a warning about unstable API since (as far as I know) the API should be stable now.
- Add a hints section for those coming from ros(1)_control.
- Reduce wordiness in a few places. It seemed like there was some redundancy, especially in stating goals of the package multiple times.
- Minor grammar cleanup.


bmagyar (approved):
It looks good to me but you may want to have a look at this repo too, I think we should try to grow docs here: https://ros-controls.github.io/control.ros.org/

PR #64, opened 2021-02-06
Adding Gazebo for rrbot

olivier-stasse:
Objective
=======
Following the nice work of @ahcorde on gazebo_ros2_control adding Gazebo for rrbot.

This pull requests modifies the rrbot_description to call the `gazebo_ros2_control`  and the `GazeboSystem`plugins.
Given the interfaces of rrbot this is directly compatible.
The rrbot_system_description_position_only.urdf.xacro has been modified to take simu has parameter.
The default is false and switch quietly to the usual behavior.
If set to true it loads the gazebo system from `gazebo_ros2_control`.

The `rrbot_gazebo.xacro` and `rrbot_system_position_only.ros2_control.xacro` are modified accordingly.

What is left ?
==========

* **Major**: Testing by sending commands to the system. Should we try to use the same forward_command_trajectory or use the same example than `gazebo_ros2_control` ? The latter might lead to an unnecessary duplication of code between the two repositories.

* **Minor**: Modifying the README to launch the test:
`ros2 launch ros2_control_demo_robot rrbot_system_position_only_gazebo.launch.py`

### DIFF
--- a/ros2_control_demo_robot/controllers/rrbot_gazebo_forward_controller_position.yaml
+++ b/ros2_control_demo_robot/controllers/rrbot_gazebo_forward_controller_position.yaml
@@ -1,0 +1,16 @@
+controller_manager:
+  ros__parameters:
+    update_rate: 100  # Hz
+
+    joint_trajectory_controller:
+      type: joint_trajectory_controller/JointTrajectoryController
+
+    joint_state_controller:
+      type: joint_state_controller/JointStateController
+
+joint_trajectory_controller:
+  ros__parameters:
+    joints:
+      - joint1
+      - joint2
+    interface_name: position
--- a/ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
+++ b/ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
@@ -1,16 +1,16 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
 
-  <xacro:macro name="rrbot_system_position_only" params="name prefix simu:='false' ">
+  <xacro:macro name="rrbot_system_position_only" params="name prefix use_sim:='false' ">
 
     <ros2_control name="${name}" type="system">
 
-      <xacro:if value="$(arg simu)">
+      <xacro:if value="$(arg use_sim)">
         <hardware>
           <plugin>gazebo_ros2_control/GazeboSystem</plugin>
         </hardware>
       </xacro:if>
-      <xacro:unless value="$(arg simu)">
+      <xacro:unless value="$(arg use_sim)">
         <hardware>
           <plugin>ros2_control_demo_hardware/RRBotSystemPositionOnlyHardware</plugin>
           <param name="example_param_hw_start_duration_sec">2.0</param>
--- a/ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
+++ b/ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
@@ -1,15 +1,24 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
 
-  <xacro:macro name="rrbot_system_position_only" params="name prefix">
+  <xacro:macro name="rrbot_system_position_only" params="name prefix simu:='false' ">
 
     <ros2_control name="${name}" type="system">
-      <hardware>
-        <plugin>ros2_control_demo_hardware/RRBotSystemPositionOnlyHardware</plugin>
-        <param name="example_param_hw_start_duration_sec">2.0</param>
-        <param name="example_param_hw_stop_duration_sec">3.0</param>
-        <param name="example_param_hw_slowdown">2.0</param>
-      </hardware>
+
+      <xacro:if value="$(arg simu)">
+        <hardware>
+          <plugin>gazebo_ros2_control/GazeboSystem</plugin>
+        </hardware>
+      </xacro:if>
+      <xacro:unless value="$(arg simu)">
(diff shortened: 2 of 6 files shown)


ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
```diff
@@ -1,15 +1,24 @@
...
+        <hardware>
+          <plugin>gazebo_ros2_control/GazeboSystem</plugin>
+        </hardware>
+      </xacro:if>
+      <xacro:unless value="$(arg simu)">
+        <hardware>
+          <plugin>ros2_control_demo_hardware/RRBotSystemPositionOnlyHardware</plugin>
+          <param name="example_param_hw_start_duration_sec">2.0</param>
+          <param name="example_param_hw_stop_duration_sec">3.0</param>
+          <param name="example_param_hw_slowdown">2.0</param>
```
ahcorde, line 18: This is not parsed in the plugin

ahcorde (changes requested):

ros2_control_demo_robot/description/rrbot/rrbot_system_position_only.ros2_control.xacro
olivier-stasse: Maybe I misunderstood your comment, but this part is not included in the file generated by xacro because it is in the `unless` statement.
Should it be included ?

ros2_control_demo_robot/launch/rrbot_system_position_only_gazebo.launch.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
olivier-stasse: Done through commit 68bdb89

ros2_control_demo_robot/package.xml
```diff
@@ -14,6 +14,7 @@
   <depend>hardware_interface</depend>
   <depend>pluginlib</depend>
   <depend>rclcpp</depend>
+  <depend>gazebo_ros2_control</depend>
```
Karsten1987, line 17: you're missing an `<exec_depend>xacro</exec_depend>` here.

ros2_control_demo_robot/launch/rrbot_system_position_only_gazebo.launch.py
```diff
@@ -0,0 +1,60 @@
...
+    node_robot_state_publisher = Node(
+        package='robot_state_publisher',
+        executable='robot_state_publisher',
+        output='screen',
+        parameters=[robot_description]
+    )
+
+    spawn_entity = Node(package='gazebo_ros', executable='spawn_entity.py',
+                        arguments=['-topic', 'robot_description',
+                                   '-entity', 'cartpole'],
```
Karsten1987, line 53: what's that cartpole doing here?

ros2_control_demo_robot/description/rrbot_system_position_only.urdf.xacro
```diff
@@ -5,7 +5,8 @@ Copied and modified from ROS1 example:
 https://github.com/ros-simulation/gazebo_ros_demos/blob/kinetic-devel/rrbot_description/urdf/rrbot.xacro
 -->
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="2dof_robot">
-
+  <xacro:arg name="simu" default="false" />
```
Karsten1987, line 8: nitpick, but `simu` sounds kinda weird. I'd propose either just `sim` or something like `use_sim`.

ros2_control_demo_robot/controllers/rrbot_gazebo_forward_controller_position.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+    joint_state_controller:
+      type: joint_state_controller/JointStateController
+
+joint_trajectory_controller:
+  ros__parameters:
+    joints:
+      - joint1
+      - joint2
+    interface_name: position
```
Karsten1987, line 16: @destogl can you explain what that interface name does here? I'd assume that the joints are sufficiently described in the ros2_control tag with their control and state interfaces names.
What if I use `position` here and declare the joints as `effort`?

ros2_control_demo_robot/package.xml
olivier-stasse: Thanks. Fixed in e6497f1e570cd3e4e2afeb1d4deeaaf6161c1bf3

ros2_control_demo_robot/launch/rrbot_system_position_only_gazebo.launch.py
olivier-stasse: Thanks. Fixed in e230567def6007d9b3bfb0c0198f964b2a8cec60

ros2_control_demo_robot/description/rrbot_system_position_only.urdf.xacro
olivier-stasse: Thanks. Fixed in 5e726d4f7000c2162c1e647d43d20339fee46f58

olivier-stasse:
I forget to mention one behavior that I found strange (or at least that I do not have with the original example from @ahcorde): gazebo takes a long time to start.

Any feedback on this part is welcome.

Karsten1987:
I've noticed the same behavior, that it takes a long time to start up.
My guess is however not that it's Gazebo's fault, but I suspect it to be xacro. I feel that instead of loading a xacro file, we should maybe try to load a URDF directly. Can you try this?
https://github.com/ros-simulation/gazebo_ros2_control/blob/master/gazebo_ros2_control_demos/launch/cart_example_position.launch.py#L40

olivier-stasse:
Superseded by #85

PR #71, opened 2021-02-22
Add corrections and extensions for the rrbot

destogl:
Replaces a part of the #53 and resolves #69 

fixes mostly trivial changes and make the example more "exemplary" :)

PR #72, opened 2021-03-02
Add test node package

destogl:
After #71 is merged, this needs a rebase.


bmagyar:
@destogl ready for rebase, once done, feel free to drop me a note to review this

destogl:
@bmagyar Before merge please test it with the manual.

ros2_control_test_nodes/test/test_copyright.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
bmagyar, line 1: we have linters for this on the CI, let's use those please, I'd really prefer if we could get rid of junk like this

ros2_control_test_nodes/ros2_control_test_nodes/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Stogl Robotics Consulting UG (haftungsbeschränkt)
```
bmagyar, line 1: do we really need to litter in the init.py files too??

ros2_control_test_nodes/resource/ros2_control_test_nodes
```diff
@@ -0,0 +1 @@
+
```
bmagyar, line 1: empty file, not intended?

bmagyar (changes requested):
please don't forget to add the new package to the gh workflows

ros2_control_test_nodes/ros2_control_test_nodes/__init__.py
destogl: Unfortunately, yes....

ros2_control_test_nodes/resource/ros2_control_test_nodes
destogl: It has to be there in ros2 python packages... No idea why. See, e.g., https://github.com/ros2/launch/blob/master/launch/resource/launch

destogl:
@bmagyar just updated... Please one more check.

bmagyar (approved):

PR #74, opened 2021-03-27
Add example 6Dbot for tests and debugging

destogl:
This is an example 6Dbot using fake components to support integration tests.

Signed-off-by: Denis Štogl <denis@stogl.de>


destogl:
closed in favor of #201

PR #77, opened 2021-04-07
Add pre-commit Checks

destogl:
This PR adds pre-commit to the repo integrating linters and formatters. This integrated all the formatters provided also by ros-tooling used in the repo:

- ament_cppcheck
- ament_cpplint
- ament_lint_cmake (added
- ament_copyright

** IMPORTANT** - replaced ROS2 ament-linters by upstream
- check-xml (replaces `ament_xmllint`)
- pep257 (replaces `ament_pep257` using the same configuration)
- flake8 (replaces `ament_flake8` using the same configuration)
- clang-format (replaces `ament_uncrustify` using the configuration from `ament_clang_format`)

**NOTE**: on the CI only *non-ament* formatters are run, because the complexity of ROS-setup in GH-Action's-config. Still they are run using the CI config from the ros-tooling group.

### List of all other added hooks

The following standard hooks from [pre-commit-hooks repo](https://github.com/pre-commit/pre-commit-hooks) are added:

- id: check-added-large-files
- id: check-ast
- id: check-case-conflict
- id: check-docstring-first
- id: check-merge-conflict
- id: check-symlinks
- id: check-xml
- id: check-yaml
- id: debug-statements
- id: end-of-file-fixer
- id: mixed-line-ending
- id: trailing-whitespace
- id: check-byte-order-marker  # Forbid UTF-8 byte-order markers

Python related-hooks:
- pyupgrade (automatically upgrades code to python 3.6+)
- black (very well python formatter)

CPP related-hooks - all menioned above.

Docs - RestructuredText hooks:
- doc8
- rst-backticks
- rst-directive-colons
- rst-inline-touching-normal
- codespell - spellcheck in the docs and comments.


Karsten1987 (commented):
I don't see how we are running the default ament linters here?
Essentially all linters specified in the linters.yaml: 
`copyright, cppcheck, cpplint, flake8, pep257, uncrustify, xmllint`

Do you mind elaborating on how these linters will run?

Karsten1987 (changes requested):
Also, please keep the things separated, you've touched 16 files - many of them because of clang-format. The exiting default linters in ament will not comply with this clang-format, i.e. uncrustify vs ament_clang_format. 

If we decide to switch to a default clang-format (which I hope we don't) we have to deactivate uncrustify for the each package in order to prevent potential test failures.

destogl:
They will not run. Because in many cases they will collide with other linters. The issue this tries to solve is flakiness of the linters from the ros-tooling.

I was following the idea from [MoveIt](https://www.youtube.com/watch?v=tOlnfrctpSY) people how to setup this.

I have personally multiple issue with default linters. Some of them are:
- `ament_copyright` fails on this PR because CONTRIBUTING.md file is not found (at least for me locally)
- `ament_uncrustify` and `ament_cpplint` are not always supporting each other (see ros-control/ros2_controllers#140)
- random failing of linters from time to time
- very inconvinient use of those and need for manual reformating (except for `ament_uncrustify`

destogl:
> If we decide to switch to a default clang-format (which I hope we don't) we have to deactivate uncrustify for the each package in order to prevent potential test failures.

Exactly this is the goal. Because it doesn't make any sense to have linters integrated into tests when the package is compiling. This makes real test just more cluttered...

For me is the same which technology we use, but it should work without thinking to much about it. This is unfortunately not the case with ament_* tools.

Karsten1987:
> They will not run. Because in many cases they will collide with other linters. The issue this tries to solve is flakiness of the linters from the ros-tooling.

I am sorry, but I don't think this is true. All the default linters are applied across the complete ROS2 core packages, none of them every being flaky. Can you point me to a PR where you can see them being contradicting across multiple runs?

> ament_copyright fails on this PR because CONTRIBUTING.md file is not found

If you look at ros2_controllers repo, I don't think it has a contribuing file either but all linters are working. Might be some other file which doesn't have a copyright notice?

> Because it doesn't make any sense to have linters integrated into tests when the package is compiling.

I don't understand this. How are the linters influencing the compilation? The linters really just invoke a tool locally in the source directory to check your files. I am pretty certain the pre-commit hooks are doing similar things. 
That being said, I am fine with having pre-commit hooks enabled, I personally really only require that all set of linters are run, specifically cppcheck, flake8 and friends. They don't care too much about code formatting, rather than missing includes or UB in python code.

destogl:
> I am sorry, but I don't think this is true. All the default linters are applied across the complete ROS2 core packages, none of them ever being flaky. Can you point me to a PR where you can see them being contradicting across multiple runs?

We has recently situation across multiple repositories that linters were failing because of some trivial bug in the upstream repos. It seems to be hard to keep stable behavior over multiple months.

> > ament_copyright fails on this PR because CONTRIBUTING.md file is not found
> 
> If you look at ros2_controllers repo, I don't think it has a contribuing file either but all linters are working. Might be some other file which doesn't have a copyright notice?

This is the output I get from the linter and the file is obviously present.

> I don't understand this. How are the linters influencing the compilation? The linters really just invoke a tool locally in the source directory to check your files. I am pretty certain the pre-commit hooks are doing similar things.
> That being said, I am fine with having pre-commit hooks enabled, I personally really only require that all set of linters are run, specifically cppcheck, flake8 and friends. They don't care too much about code formatting, rather than missing includes or UB in python code.

They are influencing functionality tests. We can think about extending these proposed tests, but we need to clean our workflow. Especially now when we are getting more traction from the community.

bmagyar:
@Karsten1987 valid points, let me try addressing them ;)

destogl:
@Karsten1987 you are right, we should not just leave out those linters. I updated the configuration with pre-commit options. Locally, on commit, all the checks will be done, but on CI ament_* tools have to be used separately because of the missing ROS environment.

What do you think?

P.S. Sorr for my reaction last week, I was very much irritated with linter-setup and overreacted.

destogl:
@Karsten1987 please check the updated descriptions with clarification of the changes.

bmagyar (approved):
looks good, I'm eager to see how this can improve workflow

bmagyar:
@destogl there's a conflict waiting to be resolve

destogl:
@bmagyar Format should fail here because no files were formatted.

PR #79, opened 2021-04-14
Reformat all files with the new pre-commit approach 

destogl:
Configuraiton is given in #77. Please discuss there.


bmagyar (approved):
The proposed changes look good to me, I think it we could start testing/refining on master

PR #85, opened 2021-04-28
Gazebo support

Karsten1987:
supersedes #64 

@olivier-stasse I hope you don't mind, but I kind of took over your PR and fixed up a bunch of things to make Gazebo actually work. With that being said, please disregard https://github.com/olivier-stasse/ros2_control_demos/pull/1
The delay we're seeing is due to the URDF being somewhat incomplete in which the description does not deliver the `hokuyo.dae` mesh file. Gazebo will thus try to find it desperately and eventually time out.
I added the mesh explicitly and set up a hook for Gazebo for load it on its model path.


Karsten1987:
I also corrected the URDF. The base link was set to `0.2m` in height, and I believe it should rather be `2m`.
When trying to load the joint state controller, one gets an exception because of name conflict. That has to do with the current duplication/renaming from `joint_state_controller` to `joint_state_broadcaster`. I've also addressed this.

olivier-stasse:
> @olivier-stasse I hope you don't mind, but I kind of took over your PR and fixed up a bunch of things to make Gazebo actually work. 
Dear @Karsten1987 not at all, thanks for fixing this and all the rest.
I hope you found #64 useful, let me know when you want me to close it.

Karsten1987:
I guess you could close your PR already, all your initial commits are still here on this PR. Would be great to have this merged at some point to have some Gazebo support upstream.

Karsten1987:
@destogl do you mind having a look at this?

bmagyar:
@Karsten1987 please mind what you have on a branch, this happened to have the contents of #89 as well which normally wouldn't be an issue but given that changes weren't cherry-picked to Olivier's branch but merged into, I couldn't just remove them by rebasing, had to opt for adding another magic blob merge commit to fix things.

bmagyar (approved):

bmagyar:
Managed to fix my gazebo setup (I think, although it's now shaky, see gif) and tested this.

Peek 2021-05-02 07-53

Merging...

PR #70, opened 2021-02-22
demo_diff_drive_controller

BorgesJVT:
Resolves #50 

- A diffbot folder was created similar to rrbot for URDF files;
- A hardware in ros2_control_demo_hardware was created with position and velocity states and velocity command;
- Launch file was added into ros2_control_demo_robot.

PR #89, opened 2021-04-28
make diff drive load correctly

Karsten1987:
fixes https://github.com/ros-controls/ros2_control_demos/pull/70#issuecomment-828716983

The launch file got also enhanced to load and start the joint state controller directly. Starting a Joint State Publisher doesn't make much sense in my opinion if you want to load the joint state controller at the same time.


bmagyar (approved):
I've fixed the cpplint warnings + ran precommit

Issue #93, opened 2021-05-13
controller manager spawner.py bug


roboticsai:
After i run the command:
`ros2 launch ros2_control_demo_robot diffbot_system.launch.py`
i'm getting the errors:
`format [py]: __init__() got an unexpected keyword argument 'parameter'`
As i debugged the source code. The bug is comming form this line:
```
            Node(
                package="controller_manager",
                executable="spawner.py",
                parameter=["joint_state_controller"],
                output="screen",
            ),
```
from given launch file. 
When i change the parameter to the arguments it works fine but still got the warnnis that controller_manager is not available. 
It shows that controller manager in ros2 has some bugs.
Any way to solve this?

Environments;
1. Os: ubuntu 20.04 desktop
2. ros: ros2 foxy

Gaurav-Kapoor-InMach:
@roboticsai I also got the same error. So i removed this node in the launch file and its was working fine for me. But there could be an error in spawner.py.

PR #94, opened 2021-05-13
correct diffdrive demo

Karsten1987:
fixes https://github.com/ros-controls/ros2_control_demos/issues/93


matthew-reynolds:
The critical change here is the `parameter=` -> `argument=`, right? Just want to verify that the deprecated `joint_state_controller` still works fine and wasn't the cause of the issue.

Karsten1987:
That is correct, even though I've seen a big fat "SEVERE WARNING" coming from pluginlib regarding name conflicts. So I've addressed that as well.

destogl (approved):
This looks good!

bmagyar (approved):
Ta!

PR #108, opened 2021-06-15
The dependency is one down

Timple:
This matters to us because we install per-package, not per git repository.


bmagyar (approved):
Correct! Nice catch!

destogl:
Thanks! But, is it? Description does not depends from hardware? Only bringup because it uses the hardware plugin. But description doesn't use any hardware plugins...

Timple:
Well the `ros2_control_demo_hardware` is used in the xacro. And if we use our own launchfiles, this whole package doesn't get installed.

It seems that the `_bringup` package only spawns 'whatever' plugins are in the xacro. It doesn't even know they depend from the package `ros2_control_demo_hardware`. It can be completely different controllers from a different package right?

destogl:
Yeah, you are right! Thanks again!

PR #122, opened 2021-07-09
Rename and document control_level vector

AndyZe:
I was a little confused about what this vector did when reading the code. Is it some kind of vector of possible control modes for each actuator? Or a vector of current control modes for all actuators? Added a comment to clarify.

I also did some minor renaming:  `lvl` --> `level` for readability and consistency.


olivier-stasse (approved):
LGM

bmagyar (approved):
thanks!

PR #127, opened 2021-07-12
Update simulation example for Gazebo and add example for Ignition

destogl:
- add new files for minimal gazebo example
- add docs how to run Gazebo example

@olivier-stasse: can you help me get this running? I am not able to do that based on the current documenation.﻿


olivier-stasse:
Sure ! Sorry for the long delay.
I'll try to give you a feedback in the following days.

olivier-stasse (changes requested):
Dear @destogl the gazebo simulation works for me.
But I could not test it because they are no default controller parameters loaded.
I guess this is just a matter of modifying `ros2_control_demo_bringup/launch/rrbot_system_position_only_gazebo.launch.py` to load the controller parameters.

Let me know if you do not have time I'll give it a try.

README.md
```diff
@@ -169,6 +169,141 @@ Controllers from this demo:
...
+   ```
+   ros2 control list_controllers
+   ```
+   You should get:
+   ```
+   joint_state_broadcaster[joint_state_broadcaster/JointStateBroadcaster] active
+   position_trajectory_controller[joint_trajectory_controller/JointTrajectoryController] active
+   ```
+
+1. If you get output from above you can send commands to *Joint Trajectory Controller*, using script:
```
olivier-stasse, line 212: I am not a native speaker but I would reformulate this sentence as:
If you get the output above you can send ...

ros2_control_demo_bringup/launch/rrbot_gazebo.launch.py
```diff
@@ -14,9 +14,8 @@
 
 from launch import LaunchDescription
 from launch.actions import IncludeLaunchDescription
-from launch.substitutions import Command, FindExecutable, PathJoinSubstitution
 from launch.launch_description_sources import PythonLaunchDescriptionSource
-
+from launch.substitutions import Command, FindExecutable, PathJoinSubstitution
```
olivier-stasse, line 18: Naive question: why not keeping the same order between the two files rrbot_gazebo.launch.py and rrbot_ignition.launch.py ?

Not very important but just making things looking the same among the examples.

ros2_control_demo_hardware/src/rrbot_with_gpios.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  return hardware_interface::return_type::OK;
+}
+
+}  // namespace ros2_control_demo_hardware
+
+#include "pluginlib/class_list_macros.hpp"
+
+PLUGINLIB_EXPORT_CLASS(
+  ros2_control_demo_hardware::RRBotSystemPositionOnlyHardware, hardware_interface::SystemInterface)
```
olivier-stasse, line 221: Apart some of the debugging messages I do not see really the difference between rrbot_with_gpios.cpp and rrbot_system_position_only.cpp ?

olivier-stasse (approved):
Dear @destogl,
Thanks a lot for the work.
I tried on Ubuntu 20.04 rolling and it is working for Gazebo.
I did not try yet on Ignition.
Apart a typo, some suggestions and the fact that I did not catch the difference between rrbot_with_gpios.cpp and rrbot_system_position_only.cpp the PR looks good to me.

ros2_control_demo_hardware/src/rrbot_with_gpios.cpp
olivier-stasse: Don't you think that the file name should be rrbot_system_with_gpios ?
And also that the class name  should be changed to RRbotSystemWithGpios ?

README.md
```diff
@@ -169,6 +169,141 @@ Controllers from this demo:
...
+   command interfaces
+        joint1/position [claimed]
+        joint2/position [claimed]
+   state interfaces
+         joint1/position
+         joint2/position
+
+   ```
+
+1. Check is controllers are running:
```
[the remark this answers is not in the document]
destogl:
```suggestion
1. Check if controllers are running:
```

README.md
destogl:
```suggestion
1. If you get the output above you can send commands to *Joint Trajectory Controller*, using script:
```

ros2_control_demo_bringup/launch/rrbot_gazebo.launch.py
destogl: Yes, we should definitely do that!

christophfroehlich:
Merging this PR into #269

PR #131, opened 2021-08-20
add control-toolbox as deb package to install

fmros:
since compilation fails without. control-toolbox installs also control_msgs which are needed from hardware_interface and are missing.


olivier-stasse (approved):
Dear @fmros thanks for the contribution !
It worked for me.
Looks good to go.

PR #132, opened 2021-08-26
Adding Example-5 from ROS2 roadmap. Work with 2 separate actuators.

bailaC:
Example - 5 : https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md#5-modular-robots-with-separate-communication-to-each-actuator


destogl (approved):
Looking good thanks for optimizing README @destogl!

README.md
```diff
@@ -391,7 +391,7 @@ Available controllers:
 - `joint_state_broadcaster[joint_state_broadcaster/JointStateBroadcaster]`
 
 Notes:
-  - Wrench messages are not displayed properly in Rviz as NaN values are not handled in Rviz and FTS Broadcaster may send NaN values.
+  - Wrench messages are maybe not displayed properly in Rviz as NaN values are not handled in Rviz and FTS Broadcaster may send NaN values.
```
bmagyar, line 394:
```suggestion
  - Wrench messages are may not be displayed properly in Rviz as NaN values are not handled in Rviz and FTS Broadcaster may send NaN values.
```
fixed the english but is this actually true? shouldn't we highlight the recent extension that was made to the rviz plugin?

ros2_control_demo_bringup/launch/rrbot_modular_actuators.launch.py
```diff
@@ -0,0 +1,63 @@
...
+def generate_launch_description():
+    # Declare arguments
+    declared_arguments = []
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "prefix",
+            default_value='""',
+            description="Prefix of the joint names, useful for \
+        multi-robot setup. If changed than also joint names in the controllers' configuration \
+        have to be updated.",
```
bmagyar, line 30:
```suggestion
        multi-robot setup. If changed, joint names in the controllers' configuration \
        also have to be updated.",
```

ros2_control_demo_bringup/launch/rrbot_modular_actuators.launch.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2021 Department of Engineering Cybernetics, NTNU.
```
bmagyar, line 1: do we need to update this?

ros2_control_demo_hardware/ros2_control_demo_hardware.xml
```diff
@@ -27,6 +27,13 @@
       The ros2_control RRBot example using a system hardware interface-type and only one command and state interface. This example is the starting point to implement a hardware interface for robots with sensors providing one communication path.
     </description>
   </class>
+  <class name="ros2_control_demo_hardware/RRBotModularJoint"
+         type="ros2_control_demo_hardware::RRBotModularJoint"
+         base_class_type="hardware_interface::ActuatorInterface">
+    <description>
+      The ROS2 Control robot with 2 separate actuators which can be controlled by 2 different endpoints.
```
bmagyar, line 34:
```suggestion
      The ros2_control robot with 2 separate actuators which can be controlled by 2 different endpoints.
```
we should lead by example :D

ros2_control_demo_hardware/src/rrbot_actuator.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+std::vector<hardware_interface::StateInterface> RRBotModularJoint::export_state_interfaces()
+{
+  std::vector<hardware_interface::StateInterface> state_interfaces;
+
+  state_interfaces.emplace_back(hardware_interface::StateInterface(
+    info_.joints[0].name, hardware_interface::HW_IF_POSITION, &hw_joint_state_));
+
+  return state_interfaces;
```
bmagyar, line 94:
```suggestion
  return {hardware_interface::StateInterface(
    info_.joints[0].name, hardware_interface::HW_IF_POSITION, &hw_joint_state_)};
```

could something like this work?

bmagyar (changes requested):

ros2_control_demo_hardware/src/rrbot_actuator.cpp
destogl: It could, but I don't know... It's example... having more, clearer code is not so bad idea I think.

PR #120, opened 2021-07-05
Adding Example-4: Adding robot with individual sensor

destogl:

PR #133, opened 2021-09-01
Add Mutli-Robot Lifecycle example

destogl:
Note: very much work in progress. it will change few times in the next few weeks. PR servs to demonstrate ideas and hopefully get some early comments.

The PR contains also changes from #120 because they are needed for example. It should be rebaased/cherry-picked after merge.

There are also two PRs containing Bugfixes which will be also submitted separately.

The functionality for this PR is implemented in [`destogl/ros2-control-extensions-2` branch](https://github.com/destogl/ros2_control/tree/ros2-control-extensions-2).


v-lopez:
Very interesting, if I understand correctly (could not run it) you have 3 robots controlled by a single controller manager and using a joint_states topic, but each robot with a different prefix?

To be honest I'm not aware of the ROS2 best practices in multi robot, I'd have thought that you'd use separate controller managers in different namespaces for each robot. 

Although this I guess allows you to send trajectories spanning multiple robots, is that the purpose? Could they be integrated into a single System then?

destogl:
> Very interesting, if I understand correctly (could not run it) you have 3 robots controlled by a single controller manager and using a joint_states topic, but each robot with a different prefix?

Basically, yes.

> To be honest I'm not aware of the ROS2 best practices in multi robot, I'd have thought that you'd use separate controller managers in different namespaces for each robot.

Actually, this is only an example. Actually, use-case would be tool-swapping. I wanted to avoid implementing much in terms of new hardware interfaces so I reused the existing code. The ideas is to demonstrate Life cycle of Hardware.

> Although this I guess allows you to send trajectories spanning multiple robots, is that the purpose? Could they be integrated into a single System then?

The idea is to have different multiple-hardware running inside one `controller_manager`. For example a manipulator and multiple tools and sensors.

christophfroehlich:
Closing this one in favor of #187

PR #136, opened 2021-09-21
Extending demo tests

livanov93:
PR info:
- Depends on ros-controls/ros2_control#540
- Add action test for jtc (velocity and position)
- Change parameters for `forward_velocity_controller` publisher
- Introduces use of `disable_commands` parameter for `GenericSystem` for simulating disconnected driver
- Add extended paramterization for JTCs


olivier-stasse:
I propose to close this PR as it is clearly outdated wrt to the recent work.

christophfroehlich:
> I propose to close this PR as it is clearly outdated wrt to the recent work.

You are right. However, the disable command demo would be fine to include somewhere else.

PR #147, opened 2021-12-20
[Example 14] Modular robot with actuators not providing states, with additional sensors

destogl:
Adding example 6 where actuatore does not provides state but separate sensors are used for it.

The "mechanical connection" is implemented using socket communication.

This is the last example for differnet architectures. The rest is moresty different type combination that need further development of the framework.
﻿


README.md
```diff
@@ -492,6 +492,30 @@ Available controllers:
 Commanding the robot: see the commands below.
 
 
+### Example 6: "Modular Robots with actuators not providing states and with additional sensors"
```
jackcenter, line 495: I'm not sure if it's worth mentioning the `slowdown` here? The example sends a '5' command to the velocity controller but it spins much slower (due to a default slowdown of 50)  which may or may not be confusing to people.

jackcenter (approved):
The demo worked as expected after following the commands and all of the expected interfaces/controllers come up. I think there is a typo under the update to example 3, but the link works regardless.

ros2_control_demo_bringup/launch/rrbot_modular_actuators_without_feedback_sensors_for_position_feedback.launch.py
```diff
@@ -0,0 +1,63 @@
...
+    slowdown = LaunchConfiguration("slowdown")
+    robot_controller = LaunchConfiguration("robot_controller")
+
+    base_launch = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource([ThisLaunchFileDir(), "/rrbot_base.launch.py"]),
+        launch_arguments={
+            "controllers_file": "rrbot_modular_actuators_without_feedback_sensors_for_position_feedback.yaml",
+            "description_file": "rrbot_modular_actuators_without_feedback_sensors_for_position_feedback.urdf.xacro",
+            "prefix": prefix,
+            "use_fake_hardware": "false",
```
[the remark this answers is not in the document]
destogl: There is actually also an simple implementation of an example hardware. But yes, except output in terminal id does not matter if fake or "real" hardware is used.

README.md
destogl: Slowdown should just make the movement "smoother", i.e., mock dynamics. Otherwise, robot would jump to the goal immediately.

DasRoteSkelett:
I looked at the C++ part and it looks good to me, as parts of the code, that are not optimal (i.e. missing error checks) are clearly marked as not suited for production.

DasRoteSkelett (approved):
I looked at the C++ part and it looks good to me, as parts of the code, that are not optimal (i.e. missing error checks) are clearly marked as not suited for production.

destogl:
@DasRoteSkelett did you try to execute the example according to instructions? Did it work?

>  (i.e. missing error checks) are clearly marked as not suited for production.

This is the repository with examples, so we will likely not add any tests for this code.

christophfroehlich:
Closing in favor of #334

PR #148, opened 2021-12-30
remove invalid diff_drive_controller parameters

wmmc88:
* `publish_cmd` seems to be replaced with `publish_limited_velocity`
* `publish_wheel_data` seems to be not ported from ros1
* `preserve_turning_radius` seems to have never merged into ros1 and doesn't exist in ros2 either

Just ran into this as I was using it as reference for what parameters are available to play with. Is there some source of documentation to look for these things(like http://wiki.ros.org/diff_drive_controller#Parameters in ros1)?


bmagyar (approved):
Nice catch!! Thanks

PR #150, opened 2022-02-09
Move test nodes to ros2_controllers repository.

destogl:
Moving the test nodes to ros2_controllers repository so they can be easily relased.
Also the test nodes are actually testing controllers so the pacakge is also renamed to "ros2_controllers_test_nodes".
﻿


christophfroehlich:
This was already done by #232 in master, I changed this to galactic

destogl:
let's just close this. We don't need to care about `galactic` anymore at all. So I'll close this.

PR #152, opened 2022-03-09
Bugfix: Do not ignore prefix argument for joint names.

destogl:
This PR fixes a bug where prefixes of joint names are ignored in URDF which causes problem when using prefix, e.g., when running multiple instances of hardware.﻿

P.S. CI is failing because it has to be updated after transition to jammy. So, I would like to proceed with this anyway.

### DIFF
--- a/ros2_control_demo_description/rrbot_description/ros2_control/external_rrbot_force_torque_sensor.ros2_control.xacro
+++ b/ros2_control_demo_description/rrbot_description/ros2_control/external_rrbot_force_torque_sensor.ros2_control.xacro
@@ -18,7 +18,7 @@
         </xacro:unless>
       </hardware>
 
-      <sensor name="tcp_fts_sensor">
+      <sensor name="${prefix}tcp_fts_sensor">
         <state_interface name="force.x"/>
         <state_interface name="force.y"/>
         <state_interface name="force.z"/>
--- a/ros2_control_demo_description/rrbot_description/ros2_control/rrbot_modular_actuators.ros2_control.xacro
+++ b/ros2_control_demo_description/rrbot_description/ros2_control/rrbot_modular_actuators.ros2_control.xacro
@@ -10,7 +10,7 @@
         <param name="example_param_hw_stop_duration_sec">3.0</param>
         <param name="example_param_hw_slowdown">${slowdown}</param>
       </hardware>
-      <joint name="joint1">
+      <joint name="${prefix}joint1">
         <command_interface name="position">
           <param name="min">-100</param>
           <param name="max">100</param>
@@ -25,7 +25,7 @@
         <param name="example_param_hw_stop_duration_sec">3.0</param>
         <param name="example_param_hw_slowdown">${slowdown}</param>
       </hardware>
-      <joint name="joint2">
+      <joint name="${prefix}joint2">
         <command_interface name="position">
           <param name="min">-100</param>
           <param name="max">100</param>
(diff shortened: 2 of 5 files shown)


bmagyar (approved):

ros2_control_demo_description/rrbot_description/ros2_control/external_rrbot_force_torque_sensor.ros2_control.xacro
```diff
@@ -18,7 +18,7 @@
         </xacro:unless>
       </hardware>
 
-      <sensor name="tcp_fts_sensor">
+      <sensor name="${prefix}tcp_fts_sensor">
```
aprotyas, line 21: Nit: Shouldn't there be some delimiter (maybe an underscore) between `${prefix}` and the rest of the identifier? Not that it matters too much but I think I prefer `foo_tcp_fts_sensor` over `footcp_fts_sensor`.

jackcenter: Prefixes aren't always used, in which case if this was changed you'd get something like `_tcp_fts_sensor`. When assigning a prefix the pattern I've seen here is to add the delimiter to the prefix. Something like: `foo_`. I agree that `foo_tcp_fts_sensor` is definitely better than `footcp_fts_sensor`.

aprotyas: Good observation, in which case I'd suggest something like this:
```xml
<xacro:macro name="external_rrbot_force_torque_sensor" params="name prefix:='' use_sim:=^|false use_fake_hardware:=^|true fake_sensor_commands:=^|false">

...

<xacro:property name="prefix_" value='${prefix + "_" if prefix else ""}' />
<sensor name="${prefix_}tcp_fts_sensor">

...
```

jackcenter (approved):
There were some joints without prefixs in: `rrbot_modular_actuators.ros2_control.xacro` and `[rrbot_system_multi_interface.ros2_control.xacro`. Should those get added there too?

ros2_control_demo_description/rrbot_description/ros2_control/external_rrbot_force_torque_sensor.ros2_control.xacro
jackcenter: That is a nice way to handle that. I guess it's up to the owners if they would like to see that updated!

destogl: @aprotyas this is a very good idea! And I was using this before. Nevertheless, I realized that having implicit changes is confusing for users. In the current way the `prefix` is very explicit, and it will be directly applied to the names. You can always as whatever delimiter you like in your launch files. 

Simple answer: "explicity before implicity" (IMHO it's easier to debug)

destogl:
> There were some joints without prefixs in: `rrbot_modular_actuators.ros2_control.xacro` and `[rrbot_system_multi_interface.ros2_control.xacro`. Should those get added there too?

oh yes definitely. Will follow up here...

ros2_control_demo_description/rrbot_description/ros2_control/external_rrbot_force_torque_sensor.ros2_control.xacro
aprotyas: I like that thought. Is there a way to document this behavior anywhere, maybe in the README?

PR #155, opened 2022-03-10
fixing command line in Step 5 in diffbot example

HusamZain:
In ReadMe file for diffbot example step 5, it says 'If everything is fine, now you can send a command to Diff Drive Controller using ros2 cli interface:'
`ros2 topic pub --rate 30 /cmd_vel geometry_msgs/msg/Twist "linear:
 x: 0.7
 y: 0.0
 z: 0.0
angular:
 x: 0.0
 y: 0.0
 z: 1.0"`
where it should be 
`ros2 topic pub --rate 30 /diffbot_base_controller/cmd_vel_unstamped geometry_msgs/msg/Twist "linear:
 x: 0.7
 y: 0.0
 z: 0.0
angular:
 x: 0.0
 y: 0.0
 z: 1.0"`


progtologist (changes requested):
Instead of changing the readme, it would be better if you changed the launch file which is actually the location where the problem gets generated [here](https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_bringup/launch/diffbot.launch.py#L63).

Briancbn:
> Instead of changing the readme, it would be better if you changed the launch file which is actually the location where the problem gets generated [here](https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_bringup/launch/diffbot.launch.py#L63).

I think having a namespace here is valid, especially under multi robot use case, this could be a good example.

However, I am not too sure the global name remapping should be done by `robot_state_publisher`.

Briancbn (approved):
lgtm, since it should solve the issue.

PR #176, opened 2022-05-03
Fix the open_loop flag in diffbot_controllers.yaml

### LINKED ISSUE #175: open_loop flag is used differently in rrbot vs diffbot
`open_loop_control: true` vs `open_loop: true`

https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_bringup/config/rrbot_controllers.yaml#L39
https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_bringup/config/diffbot_controllers.yaml

AndyZe:
Fixes #175

### DIFF
--- a/ros2_control_demo_bringup/config/diffbot_controllers.yaml
+++ b/ros2_control_demo_bringup/config/diffbot_controllers.yaml
@@ -27,7 +27,7 @@ diffbot_base_controller:
     pose_covariance_diagonal : [0.001, 0.001, 0.001, 0.001, 0.001, 0.01]
     twist_covariance_diagonal: [0.001, 0.001, 0.001, 0.001, 0.001, 0.01]
 
-    open_loop: true
+    open_loop_control: true
     enable_odom_tf: true
 
     cmd_vel_timeout: 0.5


arne48 (changes requested):
Thank you very much for pointing on this inconsistent naming.
However, after checking the controllers used in the examples it appears that a simple parameter name change in the .yaml file wouldn't suffice to achieve your goal.

You plan to use the parameter name "open_loop_control" also for the differential drive controller by changing it in this config:
[https://github.com/ros-controls/ros2_control_demos/blob/c02942a892dde71f9e2df481b8007ef7f5e173fb/ros2_control_demo_bringup/config/diffbot_controllers.yaml#L30](https://github.com/ros-controls/ros2_control_demos/blob/c02942a892dde71f9e2df481b8007ef7f5e173fb/ros2_control_demo_bringup/config/diffbot_controllers.yaml#L30)

based on the config of the rrbot:
[https://github.com/ros-controls/ros2_control_demos/blob/c02942a892dde71f9e2df481b8007ef7f5e173fb/ros2_control_demo_bringup/config/rrbot_controllers.yaml#L39](https://github.com/ros-controls/ros2_control_demos/blob/c02942a892dde71f9e2df481b8007ef7f5e173fb/ros2_control_demo_bringup/config/rrbot_controllers.yaml#L39)

But these two parameters are for two different controllers.
In the case of your change the diffbot it is for the [diff_drive_controller](https://github.com/ros-controls/ros2_controllers/tree/master/diff_drive_controller).
The config you took it from is for the [joint_trajectory_controller](https://github.com/ros-controls/ros2_controllers/tree/master/joint_trajectory_controller)

Both controllers offer an open loop option, however they use different parameter names which they inquire.

diff_drive_controller:
[https://github.com/ros-controls/ros2_controllers/blob/8bbccf27c1e936a8012b067818a2a0f909c40605/diff_drive_controller/src/diff_drive_controller.cpp#L353](https://github.com/ros-controls/ros2_controllers/blob/8bbccf27c1e936a8012b067818a2a0f909c40605/diff_drive_controller/src/diff_drive_controller.cpp#L353)
looks for _open_loop_.

and joint_trajectory_controller:
[https://github.com/ros-controls/ros2_controllers/blob/8bbccf27c1e936a8012b067818a2a0f909c40605/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L673](https://github.com/ros-controls/ros2_controllers/blob/8bbccf27c1e936a8012b067818a2a0f909c40605/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L673)
looks for _open_loop_control_.

So if you now just change the parameter name in the config it will never be used and the controller will fallback on its default value because the parameter it actually looks for is not defined.

The intention is really good and should be followed up on to look for parameter name discrepancies in all controllers and unify them.

At that this PR could be approved but in the current state as I understand the situation it would actually cause the opposite of your intention and break the demo at least in its utility to act as a reference.

AndyZe:
Good point. I'll just close this.

PR #183, opened 2022-05-23
Fix warning about "output" from launch files

AndyZe:
Fix this warning due to a recent ROS2 Rolling update:

`[ERROR] [launch]: Caught exception in launch (see debug for traceback): stdoutstderr is not a valid standard output config i.e. "screen", "log" or "both"`


bmagyar (approved):
Better this way anyways

PR #190, opened 2022-07-06
Rename position_trajectory_controller to joint_trajectory_controller

haider8645:
Hello,

I am learning the ros2 control framework and found that there was a discrepancy in the naming consistency for the joint_trajectory_controller. To add clarity, I have renamed the controller to joint_trajectory_controller which fits well with the publisher and actual controller implementation names.

Additionally, I changed the spawner argument from -c to --controller-manager because these arguments were used interchangeably and was confusing for the learning. I guess the --controller-manager is more for a new learner..

Best,
Haider

### DIFF
--- a/ros2_control_demo_bringup/config/rrbot_controllers.yaml
+++ b/ros2_control_demo_bringup/config/rrbot_controllers.yaml
@@ -8,7 +8,7 @@ controller_manager:
     forward_position_controller:
       type: forward_command_controller/ForwardCommandController
 
-    position_trajectory_controller:
+    joint_trajectory_controller:
       type: joint_trajectory_controller/JointTrajectoryController
 
 
@@ -20,7 +20,7 @@ forward_position_controller:
     interface_name: position
 
 
-position_trajectory_controller:
+joint_trajectory_controller:
   ros__parameters:
     joints:
       - joint1
--- a/ros2_control_demo_bringup/config/rrbot_controllers.yaml
+++ b/ros2_control_demo_bringup/config/rrbot_controllers.yaml
@@ -8,7 +8,7 @@ controller_manager:
     forward_position_controller:
       type: forward_command_controller/ForwardCommandController
 
-    joint_trajectory_controller:
+    joint_trajectory_position_controller:
       type: joint_trajectory_controller/JointTrajectoryController
 
 
@@ -20,7 +20,7 @@ forward_position_controller:
     interface_name: position
 
 
-joint_trajectory_controller:
+joint_trajectory_position_controller:
   ros__parameters:
     joints:
       - joint1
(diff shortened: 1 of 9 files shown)


ros2_control_demo_bringup/launch/rrbot.launch.py
```diff
@@ -80,7 +80,7 @@ def generate_launch_description():
     robot_controller_spawner = Node(
         package="controller_manager",
         executable="spawner.py",
-        arguments=["forward_position_controller", "-c", "/controller_manager"],
+        arguments=["joint_trajectory_controller", "--controller-manager", "/controller_manager"],
```
destogl, line 83: There is no need to change this line. We actually want to show here how one can switch controllers. Therefore, default started controller is a forwarding controller.

```suggestion
        arguments=["forward_position_controller", "--controller-manager", "/controller_manager"],
```

ros2_control_demo_bringup/config/rrbot_controllers.yaml
```diff
@@ -8,7 +8,7 @@ controller_manager:
     forward_position_controller:
       type: forward_command_controller/ForwardCommandController
 
-    position_trajectory_controller:
+    joint_trajectory_controller:
```
destogl, line 11: What if we call it "joint_trajectory_position_controllers"?

There are multiple versions of trajectory controller, depending on which interfaces they are controlling.

destogl (commented):
Few comments to clarify.

ros2_control_demo_bringup/launch/rrbot.launch.py
haider8645: Ok I understand. I think that if only one type of controller is used in the example then it will be easier for the user/learner of ros2 control to follow the intent. After this example another example can be used to showcase the CLI capabilities regarding the loading/unloading of another type of controller. 

We can leave both controllers in the controllers.yml and start with the joint_trajectory_controller and the switching can be done to the forward_position_controller. This will fit well to the name change to "joint_trajectory_position_controller" as suggested in the second comment. What do you think?

ros2_control_demo_bringup/config/rrbot_controllers.yaml
haider8645: I think that joint_trajectory_position_controller is a perfect fit to the intent behind the example and the name of the controller+interface usage 

I will make the necessary changes on Monday

ros2_control_demo_bringup/launch/rrbot.launch.py
haider8645: @destogl updated everything. By default, the forward_position_controller is started.

destogl (approved):
Thank you very much for your effort and the great job you did!!! 🚀

destogl:
We are generally doing PR against master, and then we can back port this to `foxy`

destogl:
@mergifyio backport galactic

PR #194, opened 2022-07-15
Fix bug that doesn't reset commands properly on activation.

### LINKED ISSUE #178: Bug in rrbot_system_with_sensor.cpp on_activate() method
In https://github.com/ros-controls/ros2_control_demos/blob/master/ros2_control_demo_hardware/src/rrbot_system_with_sensor.cpp line 175 within the on_activate() it says:

```
hw_joint_states_[i] = hw_joint_states_[i];
```

This should be:

```
hw_joint_commands_[i] = hw_joint_states_[i];
```

destogl:
Resolves #178


bmagyar (approved):
Ouch, nice catch! :D

PR #207, opened 2022-08-26
Pr 190 merge resolved

haider8645:
Hello,

this MR resolved the mergify conflicts from https://github.com/ros-controls/ros2_control_demos/pull/198

Best,
Haider


ros2_control_demo_bringup/launch/test_joint_trajectory_position_controller.launch.py
```diff
@@ -32,8 +32,8 @@ def generate_launch_description():
         [
             Node(
                 package="ros2_control_test_nodes",
-                executable="publisher_joint_trajectory_controller",
-                name="publisher_joint_trajectory_controller",
+                executable="publisher_joint_trajectory_position_controller",
```
destogl, line 35: We should here switch to `ros2_controllers_test_nodes` from ros2_controllers repository. Can you do that, please?

destogl (changes requested):
Thanks! We should do a switch here. Can you do that?

destogl:
I am not sure if I would continue with this change. Personally, find a very nice and descriptive example that the controller name is different than the type. The controller name should be rather descriptive and not correspond to its type.

destogl:
After merging #232 this is obsolete.

PR #208, opened 2022-09-11
RRBot Demo 5b broken, update code fixed

t3ch9:


dcconner:
I posted issue #210 before I saw this PR.  To me, using the full name seems correct and most general, and the issue is remapping of topic in demo launch.

destogl (approved):

PR #215, opened 2022-12-21
Add 'publish_topic' parameter for publisher_forward_position_controler

mamueluth:
* With this you can specifiy to with topic should be published


mamueluth:
Done
https://github.com/ros-controls/ros2_controllers/pull/494

PR #218, opened 2023-01-10
🔧 Fixes and updated on pre-commit hooks and their action.

destogl:
- Use pydocstyle instead of pep257 because it replaces it.
- Bump versions of hooks.
- Remove fixed python version from formatting action.
- Fix action warnings about Node.js 12 by updating them.
- Bump version of clang-format.
- correct formatting.

### DIFF
--- a/ros2_control_test_nodes/ros2_control_test_nodes/publisher_forward_position_controller.py
+++ b/ros2_control_test_nodes/ros2_control_test_nodes/publisher_forward_position_controller.py
@@ -27,7 +27,7 @@ class PublisherForwardPosition(Node):
         self.declare_parameter("goal_names", ["pos1", "pos2"])
 
         # Read parameters
-        controller_name = self.get_parameter("controller_name").value
+        # controller_name = self.get_parameter("controller_name").value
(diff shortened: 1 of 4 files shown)


ros2_control_test_nodes/ros2_control_test_nodes/publisher_forward_position_controller.py
```diff
@@ -27,7 +27,7 @@ def __init__(self):
         self.declare_parameter("goal_names", ["pos1", "pos2"])
 
         # Read parameters
-        controller_name = self.get_parameter("controller_name").value
+        # controller_name = self.get_parameter("controller_name").value
```
bmagyar, line 30: bam! what is this sir?

destogl: I wanted to push “illegal” stuff through.

The variable is not actually used, so linter is complaining. 

We have to anyway remove this package since it is moved to the controllers.

bmagyar (approved):

PR #230, opened 2023-02-02
Use /forward_position_controller/commands for all demos

christophfroehlich:
Some of the demos did not work with forward_position_controller due to mismatch of the command topic name. 

I removed the remapping in one launch file and changed the remains to have consistent `/forward_position_controller/commands` commands for all demos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ The *RRBot* URDF files can be found in the `urdf` folder of `rrbot_description` 
 
    a. Manually using ros2 cli interface:
    ```
-   ros2 topic pub /position_commands std_msgs/msg/Float64MultiArray "data:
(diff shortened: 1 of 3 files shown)


bmagyar:
I suppose this will now conflict with the package restructure. Do you want to add these changes to those PRs in the mean time?

christophfroehlich:
> I suppose this will now conflict with the package restructure. Do you want to add these changes to those PRs in the mean time?

of course. @destogl @bmagyar  what is your opinion, is the remapping of the commands in this case some "best-practice" or "laziness"? ;)

erickisos (approved):
IMO avoiding the remappings for something that it's in the docs it's be better, this change provide a clear visibility around the topic naming, so looks great to me! 💜

PR #243, opened 2023-02-19
Integrate JTC switching demo to example_1 

### LINKED ISSUE #75: Update to new ros2controlcli interface
Replace the following:

* `ros2 control load_start_controller` to `ros2 control load --state start`
*  `ros2 control load_configure_controller` to `ros2 control load --state configure`

christophfroehlich:
also closes #75


christophfroehlich:
I realized the same misleading results as described here https://github.com/ros-controls/ros2_control/issues/786
One has to update the readme if there is a change then.

README.md
```diff
@@ -497,140 +497,3 @@ Available controllers:
 
 Moving the robot:
   - see below description of `forward_position_controller`
-
```
destogl, line 500: I would also add into the main readme that example of controller switching is under example 1. What do you think?

example_1/package.xml
```diff
@@ -19,6 +19,8 @@
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
 
+  <exec_depend>ros2_controllers_test_nodes</exec_depend>
```
destogl, line 22: Doesn't all packages automatically depend on this?

christophfroehlich: what do you mean by automatically?
I would say that lots of dependencies are actually missing here, `joint_trajectory_controller` etc. Or would you include this in the ros2_control_demo meta package? (which has to be updated as well)

christophfroehlich: yes, these dependencies are missing for all examples. I can add them to all examples once this one is merged into new structure branch?

destogl: yes please!

PR #232, opened 2023-02-02
Restructuring of the repository for better overview.

destogl:
This PR introduces the restructured version of the repository.

Each example should be submitted as separate PR to this branch.
All examples should be merged at once.

Resolves #231.

Before merging this PR we should at least do the following:

- [x] rename `use_fake_hardware` to `use_mock_hardware`
- [ ] Remove `rrbot_base.launch.py` from example folders, i.e., rename it to example launch file and setup parameters needed for the example.
- [x] #240 
- [x] remove use of **NOTE**: from .rst files and use .. note:: block
- [x] Use CMakeLists.txt structure for overriding showed in - ros_controls/ros2_control#926
- [x] Remove test_nodes package from this repository and use ros2_controllers_test_nodes package 
- [x] #75 
- [ ] Check that the `.rst` files are integrated into documentation on control.ros.org
- [x] #244 

Make sure the example implements add details as: (copy this list the PR) - replaces https://github.com/ros-controls/ros2_control_demos/issues/119

- [x] Has correct folder structure (Described in the main README.md)
- [x] Example has "README.rst" with description
- [x] There are detailed commands how to run an example
- [x] There are output examples of commands
- [x] There are screenshots with expected visualizations

PR #246, opened 2023-02-23
Cleanup new structure

christophfroehlich:
This PR closes some of the open tasks of #232


christophfroehlich:
I hope I got the CMakeLists.txt structure right, at least it compiles and the examples work.

doc/index.rst
```diff
@@ -0,0 +1,7 @@
+.. _ros2_control_demos:
+
+#################
+Demos
```
christophfroehlich, line 4: Which infos should we move from README.md to this rst file? All except build instructions and build?

If someone starts from control.ros.org with installed debian packages, the `Getting started` section is maybe not necessary in the documentation but only in the readme if the repo is checked out?

What about the figures in the doc folder, they are at least not up-to-date?

christophfroehlich: I made a suggestion, feel free to comment.
I had to rename the `README.rst` to `doc/userdoc.rst,` because it would be [excluded otherwise](https://github.com/ros-controls/control.ros.org/blob/master/conf.py#L71). Is there a reason for that?

bmagyar: Also I think rst would be upset about the too many ###?

christophfroehlich: sphinx did not complain. I think the type of character does not matter.

README.md
```diff
@@ -44,78 +37,9 @@ The concepts in this package are demonstrated on the examples of *RRBot* and *Di
...
-* In ros2_control, all parameters for the driver are specified in the URDF.
-  The ros2_control framework uses the **<ros2_control>** tag in the URDF.
-* Joint names in <ros2_control> tags in the URDF must be compatible with the controller's configuration.
-
-
 ## Build status
 
-ROS2 Distro | Branch | Build status | Documentation
+ROS 2 Distro | Branch | Build status | Documentation
 :---------: | :----: | :----------: | :-----------:
```
destogl, line 43:
```suggestion
:----------: | :----: | :----------: | :-----------:
```

destogl (approved):
Simply great!

PR #255, opened 2023-03-16
[Example 8] split hw interfaces -> state & command interfaces

muritane:
Adjusted to ros2_control commit to split interfaces: https://github.com/ros-controls/ros2_control/pull/938/files
image


destogl (approved):
Ok, so this is an updated version of the demo now after splitting the interfaces?

Thanks for this!

christophfroehlich (approved):
This works, thanks!

Issue #261, opened 2023-03-26
example_1 : 


olivier-stasse:
Dear @christophfroehlich, 
Thanks again for the great work in restructuring the repo.

After cloning ```ros2_control``` from HEAD to [c8cfa05](https://github.com/ros-controls/ros2_control/commit/c8cfa053de8a2d69d6f10e770fefbd4ff051d58a), ```ros2_controllers``` from HEAD to  82a4e24 

I tried the following:
```
ros2 launch ros2_control_demo_example_1 rrbot.launch.py
ros2 launch ros2_control_demo_example_1 test_joint_trajectory_controller.launch.py
```
and the robot is not moving. 

The output of ```rrbot.launch.py```:
```
[ros2_control_node-1] [INFO] [1679797583.074655377] [RRBotSystemPositionOnlyHardware]: Reading...
[ros2_control_node-1] [INFO] [1679797583.074724709] [RRBotSystemPositionOnlyHardware]: Got state 0.00000 for joint 0!
[ros2_control_node-1] [INFO] [1679797583.074769125] [RRBotSystemPositionOnlyHardware]: Got state 0.00000 for joint 1!
[ros2_control_node-1] [INFO] [1679797583.074802707] [RRBotSystemPositionOnlyHardware]: Joints successfully read!
[ros2_control_node-1] [INFO] [1679797583.074926205] [RRBotSystemPositionOnlyHardware]: Writing...
[ros2_control_node-1] [INFO] [1679797583.074956538] [RRBotSystemPositionOnlyHardware]: Got command 0.00000 for joint 0!
[ros2_control_node-1] [INFO] [1679797583.074993370] [RRBotSystemPositionOnlyHardware]: Got command 0.00000 for joint 1!
[ros2_control_node-1] [INFO] [1679797583.075030203] [RRBotSystemPositionOnlyHardware]: Joints successfully written!
[ros2_control_node-1] [INFO] [1679797583.175206210] [RRBotSystemPositionOnlyHardware]: Reading...
[ros2_control_node-1] [INFO] [1679797583.175299375] [RRBotSystemPositionOnlyHardware]: Got state 0.00000 for joint 0!
[ros2_control_node-1] [INFO] [1679797583.175336207] [RRBotSystemPositionOnlyHardware]: Got state 0.00000 for joint 1!
[ros2_control_node-1] [INFO] [1679797583.175373040] [RRBotSystemPositionOnlyHardware]: Joints successfully read!
[...]

PR #262, opened 2023-03-27
Add readmes in every example folder

christophfroehlich:
As suggested in #261 I added a brief readme in every example folder.

### DIFF
--- a/example_1/README.md
+++ b/example_1/README.md
@@ -1,0 +1,5 @@
+# ros2_control_demo_example_1
+
+   *RRBot* - or ''Revolute-Revolute Manipulator Robot'' - a simple position controlled robot with one hardware interface. This example also demonstrates the switching between different controllers.
+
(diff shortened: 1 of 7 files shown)


olivier-stasse (approved):
Thanks a lot @christophfroehlich !
It looks good to me.

destogl (approved):
Thanks! Great!

What are your thoughts on adding more explanation there about examples? For example explanation from here: https://github.com/ros-controls/roadmap/blob/master/design_drafts/components_architecture_and_urdf_examples.md.

Or is it better to add this under `userdoc.rst`

christophfroehlich:
Good idea to include more infos about that. I prefer packing all relevant information into the userdoc.rst only, because this is the one being published to _control.ros.org_

olivier-stasse:
I agree with @christophfroehlich we should avoid duplicating information. Indicating where the information is located would be enough.

bmagyar (approved):

PR #265, opened 2023-04-09
Fix PR template

christophfroehlich:
The PR template is never shown, maybe this is due the blank line at the beginning?

I also updated the info inside the template.


bmagyar (approved):

christophfroehlich:
Hm, I still do not see the template in action. Where/when should this appear?

PR #267, opened 2023-04-09
Remove RHEL builds from PRs

christophfroehlich:
This workflow always failed -> I copied the configuration from https://github.com/ros-controls/ros2_control/pull/983

Do we need this here at all @bmagyar?


bmagyar:
Prolly never needed it

bmagyar (approved):

PR #191, opened 2022-07-07
Update gazebo rrbot example

haider8645:
Hello,

I noticed that the example of using the rrbot in gazebo together with the example publisher needed some updates. After these changes, the controller is spawned after you load the rrbot_system_position_only_gazebo.launch.py

Best,
Haider

PR #269, opened 2023-04-09
[Example 9] Gazebo Classic and Gazebo

### LINKED ISSUE #65: Making rrbot position only working on Gazebo
- [x] Link with gazebo_ros2_control: Candidate PR #64
- [ ] Example of joint_trajectory_controller applied to gazebo_ros2_control

christophfroehlich:
This is a rewrite of existing contributions #127 and #191 for the new structure of this repo. 

- [x] Basic RRBot with position interface and hardware interface copied from example_1
- [x] Working example for Gazebo Classic
- [x] Working example for Gazebo
- [x] Polishing userdoc.rst

@destogl @olivier-stasse or @haider8645: I was not able to cherry pick commits from the existing PRs due to the structure rewrite, but I added you as co-authors.

To keep that example simple, I removed JTC from this example.

If merged this closes #65


.github/workflows/rolling-semi-binary-build.yml
```diff
@@ -12,6 +12,11 @@ on:
     # Run every morning to detect flakiness and broken dependencies
     - cron: '33 1 * * *'
 
+#TODO(christophfroehlich): remove once it is released
+# otherwise ros-gz won't build
+env:
+  GZ_VERSION: fortress
```
christophfroehlich, line 18: This does not work, the environment variable seems to be not available while while running rosdep. Is there a way to pass it?

ahcorde: When running `rosdep`, you can use `--skip-keys` and you can install manually the Gazebo version. Maybe this will help as an inspiration https://github.com/ros-controls/gz_ros2_control/blob/master/.github/workflows/ci.yaml

christophfroehlich: thanks for pointing in this direction, I'll have a look

christophfroehlich: I'd need to change the CI files, because we use the ros-industrial CI here. I hope that everything gets released at some point..

example_9/description/ros2_control/rrbot.ros2_control.xacro
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro">
+
+  <xacro:macro name="rrbot_ros2_control" params="name prefix use_gazebo_classic:=^|false use_gazebo:=^|false">
+
+    <ros2_control name="${name}" type="system">
+      <hardware>
+        <xacro:if value="${use_gazebo_classic}">
+          <plugin>gazebo_ros2_control/GazeboSystem</plugin>
```
mhubii, line 9: maybe for fortress
`<plugin>ign_ros2_control/IgnitionSystem</plugin>` 
and onwards
<plugin>gz_ros2_control/IgnitionSystem</plugin>

christophfroehlich:
@mhubii I gave up to create a simple demo for rolling, because back then there was not everything released for rolling. Maybe this changed, or we can at least make a working example for humble?
Furthermore, gazebo classic and gazebo can't be installed in the same system? Maybe two separate docker containers would do. If you have an idea, feel free to create a new PR :)

mhubii:
sorry for dropping these random comments^^, just stumbled upon your pr and it was helpful regardless @christophfroehlich for what I was trying to achieve. 

Yes it seems only one version can be installed. But for each version individually it works for humble (didn't check rolling and new gazebo). There is only an issue where the ros 2 control gazebo plugin doesn't support multiple command interfaces if I am. It mistaken.

christophfroehlich:
Normally we target everything to rolling and backport stuff then. But if there is a reason for a PR against humble, thats OK.
If you have a working setup, can you make a new PR targeting humble? The tricky thing is describing the installation,  setup of the docker, and make CI happy.

christophfroehlich:
I close this in favor of #482 -> let's keep humble+iron as it is.

PR #66, opened 2021-02-16
WIP: Add example of starting controllers with launch files

v-lopez:
Requires https://github.com/ros-controls/ros2_control/pull/310

Signed-off-by: Victor Lopez <victor.lopez@pal-robotics.com>

PR #273, opened 2023-04-18
Add example of starting controllers with launch files

christophfroehlich:
Rewrite of #66

### DIFF
--- a/example_1/doc/userdoc.rst
+++ b/example_1/doc/userdoc.rst
@@ -120,46 +120,62 @@ Tutorial steps
     ros2 topic echo /dynamic_joint_states
 
 6. Let's switch to a different controller, the ``Joint Trajectory Controller``.
-   Load the controller manually by
-
-   .. code-block:: shell
-
-    ros2 control load_controller joint_trajectory_position_controller
-
-   what should return ``Successfully loaded controller joint_trajectory_position_controller``. Check the status
-
-   .. code-block:: shell
-
-    ros2 control list_controllers
-
-   what shows you that the controller is loaded but unconfigured.
-
-   .. code-block:: shell
-
-    joint_state_broadcaster[joint_state_broadcaster/JointStateBroadcaster] active
-    forward_position_controller[forward_command_controller/ForwardCommandController] active
-    joint_trajectory_position_controller[joint_trajectory_controller/JointTrajectoryController] unconfigured
-
-   Configure the controller by setting it ``inactive`` by
-
-   .. code-block:: shell
-
-    ros2 control set_controller_state joint_trajectory_position_controller inactive
-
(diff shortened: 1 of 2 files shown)


example_1/doc/userdoc.rst
```diff
@@ -120,46 +120,62 @@ Tutorial steps
...
+   a. Load the controller manually by
 
-    ros2 control list_controllers
+      .. code-block:: shell
 
-   what shows you that the controller is loaded but unconfigured.
+        ros2 control load_controller joint_trajectory_position_controller
 
-   .. code-block:: shell
+      what should return ``Successfully loaded controller joint_trajectory_position_controller``. Check the status
```
destogl, line 137:
```suggestion
      what should return ``Successfully loaded controller joint_trajectory_position_controller``.
      Then check controllers' status
```

example_1/doc/userdoc.rst
```diff
@@ -120,46 +120,62 @@ Tutorial steps
...
-    joint_state_broadcaster[joint_state_broadcaster/JointStateBroadcaster] active
-    forward_position_controller[forward_command_controller/ForwardCommandController] active
-    joint_trajectory_position_controller[joint_trajectory_controller/JointTrajectoryController] unconfigured
+      .. code-block:: shell
 
-   Configure the controller by setting it ``inactive`` by
+        ros2 control list_controllers
 
-   .. code-block:: shell
+      what shows you that the controller is loaded but unconfigured.
```
destogl, line 143:
```suggestion
      what shows you that the controller is loaded but unconfigured. Similar output is expected:
```

destogl (changes requested):
I like the docs extension and showing how controllers are managed. Nevertheless, I am not convinced that we need this launch file. We have great CLI options, and they can do everything required, so what is the added value into having this launch file.

@v-lopez if you are still there, please tell us.

Still, @bmagyar and @christophfroehlich, what do you think about his additional option? Is there something I don't see?

christophfroehlich:
The diff view is broken here, this PR actually only adds literal c in line 167. The rest is just restructuring of the section.

I found the old PR interesting and merged it to the new structure. I thought it could be helpful for switching controller from some external interface like a GUI, or probably shows how to integrate `controller_manager.launch_utils` to a python state machine script. But because it only loads the controller but does not activate it, I agree that the benefit of this launch file is close to zero.

olivier-stasse (approved):
For a first example I believe that the most simple example is preferable.
We can give a try but I think most of the feedback we had was that simpler is better.

christophfroehlich:
I don't have any hard feelings about this PR. If you think this is not adding any value to the demos but only making them more verbose -> I'll close it.

PR #274, opened 2023-04-21
[example_2] Remove kinematics from diffbot

christophfroehlich:
The kinematics calculation was without any use in the hw interface. 

Maybe the original intention was to simulate some odometry sensors, or to perform the state integration in world coordinates?
I removed it now, and also split the command output from the state integration into the read/write method.

### DIFF
--- a/example_2/hardware/diffbot_system.cpp
+++ b/example_2/hardware/diffbot_system.cpp
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <cmath>
+#include <cstddef>
 #include <limits>
 #include <memory>
 #include <vector>
--- a/example_2/hardware/diffbot_system.cpp
+++ b/example_2/hardware/diffbot_system.cpp
@@ -35,10 +35,6 @@ hardware_interface::CallbackReturn DiffBotSystemHardware::on_init(
     return hardware_interface::CallbackReturn::ERROR;
   }
 
-  base_x_ = 0.0;
-  base_y_ = 0.0;
-  base_theta_ = 0.0;
-
   // BEGIN: This part here is for exemplary purposes - Please do not copy to your production code
   hw_start_sec_ = std::stod(info_.hardware_parameters["example_param_hw_start_duration_sec"]);
   hw_stop_sec_ = std::stod(info_.hardware_parameters["example_param_hw_stop_duration_sec"]);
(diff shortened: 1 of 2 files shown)


example_2/hardware/diffbot_system.cpp
```diff
@@ -177,37 +173,19 @@ hardware_interface::CallbackReturn DiffBotSystemHardware::on_deactivate(
 hardware_interface::return_type DiffBotSystemHardware::read(
   const rclcpp::Time & /*time*/, const rclcpp::Duration & period)
 {
-  double radius = 0.02;  // radius of the wheels
-  double dist_w = 0.1;   // distance between the wheels
-  for (uint i = 0; i < hw_commands_.size(); i++)
+  // BEGIN: This part here is for exemplary purposes - Please do not copy to your production code
+  for (uint i = 0; i < hw_velocities_.size(); i++)
```
aprotyas, line 177: Since you're re-writing, may I suggest
```suggestion
  for (std::size_t i = 0; i < hw_velocities_.size(); i++)
```

aprotyas (commented):
I think the kinematics calculation was included as a reference, but maintainers can speak more to it?

example_2/hardware/diffbot_system.cpp
aprotyas: *Ideally* we should use the corresponding `size_type` member type of iterable STL types (`std::string`, `std::vector`, etc.) when looping against `std::some_iterable_type<T>::size()`

```cpp
std::vector<int> numbers;
for (decltype(numbers)::size_type index {}; index < numbers.size(); ++index);
```

But IMO this is not very readable and *most* STL implementations typedef `std::vector<T>::size_type` to `std::size_t` anyways, so it's usually a safe practice to follow. Including `<cstddef>` is a small price to pay!

christophfroehlich: what about `auto` in this case? (like the other 6 for-loops in this file ;))

aprotyas: `auto` is inferred as `int` for the literal `0`. Unfortunately there's no literal for `std::size_t`, but in a previous project we've used `_z` as a user provided literal for `std::size_t`. Something like this!

```cpp
constexpr std::size_t operator "" _z ( unsigned long long n )
{
  return n;
}

static_assert(std::is_same_v<decltype(0_z), std::size_t>);

// for (auto i = 0_z; i < 10_z; ++i);
```

Come C++23, we can also use the language provided `UZ` literal for this purpose!
https://en.cppreference.com/w/cpp/language/integer_literal

Issue #239, opened 2023-02-13
"on_deactivate()" seems not called when setting the controller to be "inactive"


xibeisiber:
Hi,
I tried `diffbot.launch.py` and `rrbot_system_position_only.launch.py`. The printing information in `on_activate()` is printed out in console, but when I "ros2 control set_controller_state forward_position_controller inactive" and "ros2 control set_controller_state diffbot_base_controller inactive", the printing information in "on_deactivate()" is not printed in console.

Another question is: will `on_deactivate()` be automatically called when I terminate the startup node (press ctrl+c in` diffbot.launch.py `)?

Thanks!
Jia

christophfroehlich:
Hi,
I think this is not an issue/question of this demo-repository but of https://github.com/ros-controls/ros2_control directly. 

My experience on the termination with ctrl+c is that it is not really reproducible what happens (independent of ros2_control) ;)

destogl:
> I tried `diffbot.launch.py` and `rrbot_system_position_only.launch.py`. The printing information in `on_activate()` is printed out in console, but when I "ros2 control set_controller_state forward_position_controller inactive" and "ros2 control set_controller_state diffbot_base_controller inactive", the printing information in "on_deactivate()" is not printed in console.

Oh, this is very strange. this should definitely not happen. This part of ros2_control is well tested. I would not expect bugs there, nevertheless it is possible. I will test this when testing new restructured examples in #232  to be sure. Can you check the example one from this PR and report if you get the same behavior?

> Another question is: will `on_deactivate()` be automatically called when I terminate the startup node (press ctrl+c in`diffbot.launch.py`)?

I can not answer you in detail

xibeisiber:
> > I tried `diffbot.launch.py` and `rrbot_system_position_only.launch.py`. The printing information in `on_activate()` is printed out in console, but when I "ros2 control set_controller_state forward_position_controller inactive" and "ros2 control set_controller_state diffbot_base_controller inactive", the printing information in "on_deactivate()" is not printed in console.
> 
> Oh, this is very strange. this should definitely not happen. This part of ros2_control is well tested. I would not expect bugs there, nevertheless it is possible. I will test this when testing new restructured examples in #232 to be sure. Can you check the example one from this PR and report if you get the same behavior?
> 
> > Another question is: will `on_deactivate()` be automatically called when I terminate the startup node (press ctrl+c in`diffbot.launch.py`)?
> 
[...]

PR #277, opened 2023-04-25
adjust example 1 to use hardware_spawner.

mamueluth:
Example of how to use the new way of configuring/activating the hardware.
One should note however: Currently the hardware gets still activated by default (legacy support). Only workaround would be to add something like:
```yaml
controller_manager:
  ros__parameters:
    ......
    configure_components_on_start:
      - ""
    activate_components_on_start:
      - ""
```
to the `example_1/bringup/config/rrbot_controllers.yaml` file. Which I think is kind of ugly and miss leading. Would not do this as hardware_spawner prints successful activation anyways... (returns current state of hardware which is activated)

### DIFF
--- a/example_1/bringup/launch/rrbot.launch.py
+++ b/example_1/bringup/launch/rrbot.launch.py
@@ -56,11 +56,22 @@ def generate_launch_description():
         parameters=[robot_description, robot_controllers],
         output="both",
     )
+    hardware_activation = Node(
+        package="controller_manager",
+        executable="hardware_spawner",
+        arguments=["RRBot", "--activate"],
+    )
     robot_state_pub_node = Node(
         package="robot_state_publisher",
         executable="robot_state_publisher",
         output="both",
         parameters=[robot_description],
+    )
+    delay_robot_state_pub = RegisterEventHandler(
+        event_handler=OnProcessExit(
+            target_action=hardware_activation,
+            on_exit=[robot_state_pub_node],
+        )
     )
     rviz_node = Node(
         package="rviz2",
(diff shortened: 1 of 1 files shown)


christophfroehlich:
I'm not sure about the plans on this: will  the old way of activating still be possible in the future? If yes, i wouldn't make this detail to be the default of the very first example.

Another issue: Will this work with humble release too? if no -> we have to generate a humbe branch here.

mamueluth:
I think the plan is to adjust every example. As far as I know, the old way of activating the hardware is not going to be supported anymore in the future. I don't know when this is going to happen though...
Don't see much of a burden added to the user here. The demo stays more or less the same and the only thing to realize for a user is that there is a way of activating the hw.

I don't think back porting is planned but @destogl should know more about this. You are probably right with your concerns about the humble branch.

christophfroehlich:
I remembered #239 and tried to deactivate the hw_interface with 
``` bash
$ ros2 run controller_manager hardware_spawner DiffBot --configure
[INFO] [1682761448.258076586] [hardware_spawner]: configured component 'RRBot'. Hardware now in state: lifecycle_msgs.msg.State(id=2, label='inactive').
```
The `on_deactivate` method is called with the expected output in the terminal, but after the deactivation `read` and `write` is still called. Why? Is this an issue of this hw-interface or is this not implemented properly in ros2_control?

christophfroehlich (changes requested):
I think we should add a section on that topic in the docs as well, e.g. why and how to activate/deactivate hw components after spawning.

destogl (approved):

destogl:
> The `on_deactivate` method is called with the expected output in the terminal, but after the deactivation `read` and `write` is still called. Why? Is this an issue of this hw-interface or is this not implemented properly in ros2_control?

This is a feature for now. We don't have support for configuration interfaces yes, so some time ago, we decided to make those calls also in inactive state. The reason for this is the requirement to set some configuration interface before activation. 

This issue is related to ros-controls/roadmap#51

destogl:
> I think we should add a section on that topic in the docs as well, e.g. why and how to activate/deactivate hw components after spawning.

Yes, exactly, and also add notes about this into migration notes. I am tending toward disabling automatic activation in Iron. 
The user-base will not be huge for the moment and we have the opportunity to test it well.

christophfroehlich:
> > The `on_deactivate` method is called with the expected output in the terminal, but after the deactivation `read` and `write` is still called. Why? Is this an issue of this hw-interface or is this not implemented properly in ros2_control?
> 
> This is a feature for now. We don't have support for configuration interfaces yes, so some time ago, we decided to make those calls also in inactive state. The reason for this is the requirement to set some configuration interface before activation.
> 
> This issue is related to [ros-controls/roadmap#51](https://github.com/ros-controls/roadmap/pull/51)

I see, this makes more sense now. Maybe we should go on with the proposal from https://github.com/ros-controls/roadmap/pull/51 because with the proposed changes of the hw component lifecycle this will get to attention (and confuse) more people.

destogl:
Let simplify this:

```
controller_manager:
  ros__parameters:
    hardware_interface_state_after_start:
      not_loaded:
        - ""
      unconfigured:
        - "arm1"
        - "arm2"
      inactive:
        - ""
```

destogl:
you are right! We should add this to discussion agenda and make this clear

destogl:
we add this functionality in the multi robot examples #334

PR #278, opened 2023-04-27
Add example how to use controller-specific update rate.

destogl:
@christophfroehlich what do you think is the best place to show-case this?


christophfroehlich:
I'm not sure if we should include all such small details in example1, like we have already with switching controllers etc. Could get a bit packed then?! But creating new examples for this only also makes no sense.
I suggest: 
Let's do this in example 1 for now, and split it later if it gets too big!

olivier-stasse:
@destogl, @christophfroehlich I guess we should add some explanations in the userdoc.rst to briefly explain the update_rate change ?

bmagyar (approved):

PR #297, opened 2023-06-04
Remove outdated state_publish_rate for JTC

christophfroehlich:
`state_publish_rate` was removed by https://github.com/ros-controls/ros2_controllers/pull/520/files


destogl (approved):
Thanks!

PR #303, opened 2023-06-12
Add CI workflow to check docs

christophfroehlich:
This should add a CI job with the reusable workflow from https://github.com/ros-controls/control.ros.org/pull/125

Because the workflow is not yet merged, I had to use my fork -> I'll change that before the merge.


.github/workflows/ci-check-docs.yml
```diff
@@ -0,0 +1,12 @@
+name: Check Docs
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  check-docs:
+    name: Check Docs
+    uses: christophfroehlich/control.ros.org/.github/workflows/reusable-sphinx-check-single-version.yml@reusable_wf
```
christophfroehlich, line 10:
```suggestion
    uses: ros-controls/control.ros.org/.github/workflows/reusable-sphinx-check-single-version.yml@master
```

olivier-stasse (dismissed):
LGTM

olivier-stasse (approved):
Thanks for the change.

.github/workflows/ci-check-docs.yml
christophfroehlich: I did already yesterday -> now merged into master

PR #304, opened 2023-06-15
Adding docker

ARK3r:
There is about 1 week of logs that can be seen [here](https://github.com/ros-controls/ros2_control_demos/pull/298) however there are many unnecessary commits there (about 30-something) which are not exactly informative to track, so this is just a cleaned-up version of that pull request all in one commit.


christophfroehlich:
This is not necessary, because all commits get squashed for the merge of the PR ;)

PR #305, opened 2023-06-16
add a dependency; fix deprecation warning

ARK3r:
getting rid of this warning:
https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L160C1-L167

and joint_state_publisher_gui wasn't installed when running view_robot.launch.py inside docker.

### DIFF
--- a/example_1/bringup/launch/rrbot.launch.py
+++ b/example_1/bringup/launch/rrbot.launch.py
@@ -53,7 +53,8 @@ def generate_launch_description():
     control_node = Node(
         package="controller_manager",
         executable="ros2_control_node",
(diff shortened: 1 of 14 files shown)


christophfroehlich (approved):
Thanks for both!

christophfroehlich:
@mergifyio backport humble

christophfroehlich:
I realized now that this change was not released yet with ros2_control -> our demos do not work with a ros2_control binary installation at the moment.

christophfroehlich:
@mergifyio backport humble

PR #309, opened 2023-06-18
Move installation instructions to RST and update docker usage instructions

christophfroehlich:
A follow-up on #298: I moved the instructions from @ARK3r to the rst file, which is published on control.ros.org and extended it with instructions for a local manual installation.


christophfroehlich:
@ARK3r Do you have a suggestion how we can fix the `view_robot.launch.py` to run from the docker image? joint_state_publisher_gui cannot be launched obviously. 
we could add to the instructions, that one has to install ROS locally (already necessary for rviz2) use `sudo apt-get install ros-humble-joint-state-publisher-gui`, and `run ros2 run joint_state_publisher_gui joint_state_publisher_gui`
But then we need a `gui:=false` flag for these launch files.

ARK3r:
I think doing as you suggested makes sense.

I'll try to add the flag tonight.

ARK3r:
@christophfroehlich I started adding the `gui` arg then I realised we'd need to have both rviz and joint_state_publisher_gui to run if this arg is true.

But if we want to run it in docker, then we can't have either of them run their gui.

So then we'd need to run rviz and joint_state_publisher_gui outside the docker.

Since we don't want the user to colcon build the package (because in the case that they run the build then what's the point of having the docker?), then the user would have to run rviz and joint_state_publisher_gui in two seperate terminals and I think at this point this is just too messy for the user.

Do you think there's a way we could do this cleaner?

christophfroehlich:
Exactly, we'd need both to run outside of the docker. Opening a new terminal is less messy for a beginner than having a docker where half of the examples don't work ;) We could also provide a docker with X11 forwarding, but that's not easy to work on every host platform. 
I'd go for a single `gui` argument (default true), if set to false then neither rviz2 nor jsp_gui should be launched.

ARK3r:
Made a pr to your fork: https://github.com/christophfroehlich/ros2_control_demos/pull/3

I'll add some documentation now too.

ARK3r:
I still think this is "too much work" for a new commer just to get to view the robot, and maybe we should push "view robot by docker" lower into the documentation, or maybe make a table that has an overview of the repo in one screen size so anyone coming in can see at one glance what are all the things the repo does.

christophfroehlich:
> I still think this is "too much work" for a new commer just to get to view the robot, and maybe we should push "view robot by docker" lower into the documentation, or maybe make a table that has an overview of the repo in one screen size so anyone coming in can see at one glance what are all the things the repo does.

I agree that the barriers of entry are higher with using the docker, but let us think of a necessary educational aspect ;) Let us make an easy readable introduction to the docker and a small hint on top of every example-rst, that the following commands are written for a local installation and if used with docker one has to use the prefix `docker etc`, like 

```rst
.. note::

   The following commands are written for a local installation. If you use the provided docker container, add the following prefix

   .. code::

   docker run -it --rm --name ros2_control_demos --net host ros2_control_demos  

   For more  information on the docker usage see :ref:`ros2_control_demos_install`
```

ARK3r:
is this sufficient for the rst for now?

https://github.com/christophfroehlich/ros2_control_demos/pull/5

ARK3r (approved):

ARK3r (approved):

example_9/doc/userdoc.rst
```diff
@@ -14,20 +14,19 @@ Gazebo Classic is used for this purpose.
   Gazebo Classic should be automatically installed.
 
   * If you have installed and compiled this repository locally, you can directly use the commands below.
-  * If you have installed it via the provided docker image, use the commands as described with :ref:`ros2_control_demos_install`
-    to run the examples without Gazebo Classic. To run the example with Gazebo Classic execute
+  * If you have installed it via the provided docker image: To run the first two steps of this example (without Gazebo Classic), use the   commands as described with :ref:`ros2_control_demos_install`. To run the later steps using Gazebo Classic, execute
```
ARK3r, line 17:
```suggestion
  * If you have installed it via the provided docker image: To run the first two steps of this example (without Gazebo Classic), use the commands as described with :ref:`ros2_control_demos_install`. To run the later steps using Gazebo Classic, execute
```

ARK3r (changes requested):

ARK3r (approved):

olivier-stasse (approved):
LGTM

christophfroehlich:
@mergifyio backport humble

PR #316, opened 2023-06-20
[Example 11] Carlike Robot using Bicycle Steering Controller

ARK3r:
Example with 2 front wheels and 2 rear wheels.

Copying a lot from DiffBot, I would say the only difference is that CarlikeBot uses Ackermann Steering Controller, everything else should be quite similar.

I've also added Carlikebot to [ros2_control_demo_description](https://github.com/ros-controls/ros2_control_demos/tree/master/ros2_control_demo_description) as per the new convention for the robot models.


ARK3r:
After running `carlikebot.launch.py` I get the following:
```bash
TF_NAN_INPUT: Ignoring transform for child_frame_id "rear_left_wheel" from authority "Authority undetectable" because of a nan value in the transform (nan nan nan) (nan nan nan nan)
[rviz2-5]          at line 235 in ./src/buffer_core.cpp
[rviz2-5] Error:   TF_DENORMALIZED_QUATERNION: Ignoring transform for child_frame_id "rear_left_wheel" from authority "Authority undetectable" because of an invalid quaternion in the transform (nan nan nan nan)
[rviz2-5]          at line 254 in ./src/buffer_core.cpp
[rviz2-5] Error:   TF_NAN_INPUT: Ignoring transform for child_frame_id "rear_right_wheel" from authority "Authority undetectable" because of a nan value in the transform (nan nan nan) (nan nan nan nan)
[rviz2-5]          at line 235 in ./src/buffer_core.cpp
[rviz2-5] Error:   TF_DENORMALIZED_QUATERNION: Ignoring transform for child_frame_id "rear_right_wheel" from authority "Authority undetectable" because of an invalid quaternion in the transform (nan nan nan nan)
[rviz2-5]          at line 254 in ./src/buffer_core.cpp
```

And these are the outputs from the launch file.

```bash
reza@3d65c5cee69c:/home/ws$ ros2 control list_hardware_interfaces 
command interfaces
        ackermann_steering_controller/angular/position [unavailable] [unclaimed]
        ackermann_steering_controller/linear/velocity [unavailable] [unclaimed]
        left_steering_joint/position [available] [unclaimed]
        rear_left_wheel_joint/velocity [available] [unclaimed]
        rear_right_wheel_joint/velocity [available] [unclaimed]
        right_steering_joint/position [available] [unclaimed]
state interfaces
        left_steering_joint/position
        rear_left_wheel_joint/velocity
        rear_right_wheel_joint/velocity
        right_steering_joint/position
reza@3d65c5cee69c:/home/ws$ ros2 control list_controllers 
joint_state_broadcaster[joint_state_broadcaster/JointStateBroadcaster] active    
ackermann_steering_controller[ackermann_steering_controller/AckermannSteeringController] inactive  
```

I've cross-checked this package with the Diffbot package (example_2) and I can't figure out where the issue is coming from.

christophfroehlich:
The problem is that the joint_states do not have finite values
```
$ ros2 topic echo /joint_states
header:
  stamp:
    sec: 1688117758
    nanosec: 551043008
  frame_id: ''
name:
- left_steering_joint
- right_steering_joint
- rear_left_wheel_joint
- rear_right_wheel_joint
position:
- 0.0
- 0.0
- .nan
- .nan
velocity:
- .nan
- .nan
- 0.0
- 0.0
effort:
- .nan
- .nan
- .nan
- .nan
```

ARK3r:
Thank you that solved it.

Does this mean that if a motor doesn't have positional reading, it is mandatory for the data to be augmented (by calculating speed * time)?

christophfroehlich:
Yes and no. There has to be a valid position value in the joint_states message, but if you don't care that the wheels aren't rotating in rviz2 then you also can send zeros.

christophfroehlich:
@ARK3r are there any updates on this, or are you waiting on feedback?

ARK3r:
@christophfroehlich it got pushed to the side, I'll try to wrap it up soon. 

If you could take a look at the structure of it all, next I'll add some docs.

ARK3r:
Comparing the behaviour of DiffBot and CarlikeBot I'm seeing some differences and I'm wondering if this is coming from the controllers or the configuration.

DiffBot
DiffBot

CarlikeBot:
CarlikeBot

christophfroehlich:
@ARK3r the example looks fine and works now (I just made small fixes to simplify the launch file), but I also observe the strange behavior of the controller. With twist with smaller magnitude it starts fine but ends in a spiral instead of a circle.

example_11/hardware/carlikebot_system.cpp
```diff
@@ -0,0 +1,315 @@
...
+  RCLCPP_INFO(rclcpp::get_logger("CarlikeBotSystemHardware"), "Writing...");
+
+  for (auto i = 0u; i < hw_commands_.size(); ++i)
+  {
+    hw_velocities_[i] = hw_commands_[i];
+
+    RCLCPP_INFO(
+      rclcpp::get_logger("CarlikeBotSystemHardware"), "Got command %.5f for '%s'", hw_commands_[i],
+      info_.joints[i].name.c_str());
+  }
```
christophfroehlich, line 306: I think the problem comes from this part of the code: steering command is position (angle), and not a velocity.

christophfroehlich (changes requested):

ARK3r:
The last change I made still doesn't fix the issue, but I think it's a step closer by putting all the interfaces in one vector with type tuple<string, double, double> for joint name, state, and command.

Also by taking a look at [this](https://github.com/ros-controls/ros2_controllers/blob/master/ackermann_steering_controller/doc/userdoc.rst), I noticed that for the rear, two wheels are expected, with each accepting one command, and for steering, two wheels are expected with each taking **two** commands. Does this mean I need to add separate joints and links to each steering hinge on the front for the wheels so that the rotation of the wheels can accept a command as well?

christophfroehlich:
no it expects 4 command interfaces only as far as I can remember. I'll have a look on that somewhen the next days, maybe also fixing the bug with the open_loop mode first.

christophfroehlich:
The launch file is not working since the last commits?

I added a PR to fix the open_loop mode, maybe this helps you with debugging: https://github.com/ros-controls/ros2_controllers/pull/793

christophfroehlich (changes requested):
There is an issue in the URDF: the traction joint of the front wheels has wrong axis.
image

nice-user1:
There is some transform error with this. when running example_11 "carlikebot.launch.py", only robot body is visible and the wheels are not. I have attached the log file below. The error starts from line number 78.
[error.txt](https://github.com/ros-controls/ros2_control_demos/files/13827349/error.txt)

ARK3r:
Here I changed the carlikebot description a little so that I can use the bicycle controller on a 4-wheel vehicle. There are two virtual joints, one for steering and one for traction where carlikebot's actual wheels just mimic the movements of the virtual joints. I think this will be a good baseline for using ackermann on the same urdf with minor changes and adding a couple of arguments to the xacro and launch files.

I have yet to fix the documentation but everything should be functional so I would appreciate it if you could give it a try. I also made some design changes by defining a map for hw_interfaces_ instead of a vector. I believe this will make the code way more readable.

ARK3r:
One strange thing I'm noticing is when `linear.x=0`, whatever value `angular.z` gets the steering just snaps to $\pm \pi$ which I believe comes from [SteeringOdodmetry::get_commands](https://github.com/ros-controls/ros2_controllers/blob/humble/steering_controllers_library/src/steering_odometry.cpp#L218-L222). I'm just not sure what the reason for this is and it feels strange. Wonder if anyone has an explanation.

ARK3r:
> There is an issue in the URDF: the traction joint of the front wheels has wrong axis. image

this issue is now resolved with the new description.

christophfroehlich (approved):
> One strange thing I'm noticing is when `linear.x=0`, whatever value `angular.z` gets, the steering just snaps to ±π which I believe comes from [SteeringOdodmetry::get_commands](https://github.com/ros-controls/ros2_controllers/blob/humble/steering_controllers_library/src/steering_odometry.cpp#L218-L222). I'm just not sure what the reason for this is and it feels strange. Wonder if anyone has an explanation.

I think this is fine, because this is a singular point of the inverse kinematics. It would be also fine if the bicycle controller does not accept it, because this does not work with the traction wheel not being the steered one (but with the [tricycle model](http://localhost:8081/doc/ros2_controllers/tricycle_controller/doc/userdoc.html) it would work).



> > There is an issue in the URDF: the traction joint of the front wheels has wrong axis.  
>
> this issue is now resolved with the new description.

Perfect.

> Here I changed the carlikebot description a little so that I can use the bicycle controller on a 4-wheel vehicle. There are two virtual joints, one for steering and one for traction where carlikebot's actual wheels just mimic the movements of the virtual joints. I think this will be a good baseline for using ackermann on the same urdf with minor changes and adding a couple of arguments to the xacro and launch files.

I'm fine with that. This also is aligned with my nomenclature of https://github.com/ros-controls/ros2_controllers/pull/954 (I'd be happy about some feedback there).

> Also I added a node to relay the /tf_odometry topic to /tf which you can activate by calling `ros2 launch ros2_control_demo_example_11 carlikebot.launch.py relay_odometry_tf:=true`

I changed this to use remapping instead, to avoid a dependency on topic_tools.

christophfroehlich:
@mergifyio backport iron humble

PR #162, opened 2022-03-24
Chained controllers example

destogl:
To test the example, checkout the following branches and following dependencies:

ros-controls/ros2_control#667
ros-controls/ros2_controllers#318

Then, follow the instructions in README added in this PR.

PR #318, opened 2023-06-25
[Example 12] Controller chaining example

saikishor:
This PR is to write an example of controller chaining consolidating the PR's https://github.com/ros-controls/ros2_control_demos/pull/162 and https://github.com/ros-controls/ros2_controllers/pull/556.


example_12/controllers/src/passthrough_controller.cpp
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "passthrough_controller/passthrough_controller.hpp"
+
+namespace passthrough_controller
+{
+
```
christophfroehlich, line 19: Is there a special reason why you defined the functions in the header file, if you have already a c++ file too?

example_12/controllers/include/passthrough_controller/passthrough_controller.hpp
```diff
@@ -0,0 +1,225 @@
...
+  controller_interface::CallbackReturn on_configure(const rclcpp_lifecycle::State &) override
+  {
+    params_ = param_listener_->get_params();
+    command_interface_names_ = params_.interfaces;
+
+    if (command_interface_names_.empty())
+    {
+      RCLCPP_ERROR(get_node()->get_logger(), "'interfaces' parameter list was empty");
+      return controller_interface::CallbackReturn::ERROR;
+    }
```
christophfroehlich, line 68: Shouldn't we use the validator `not_empty<> `of generate parameter lib instead?

example_12/doc/userdoc.rst
```diff
@@ -0,0 +1,168 @@
...
+    command interfaces
+          joint1/position [available] [claimed]
+          joint1_position_controller/joint1/position [unavailable] [unclaimed]
+          joint2/position [available] [claimed]
+          joint2_position_controller/joint2/position [unavailable] [unclaimed]
+    state interfaces
+          joint1/position
+          joint2/position
+
+   At this stage the reference interfaces of controllers are listed under ``command_interfaces`` when ``ros2 control list_hardware_interfaces`` command is executed.
```
christophfroehlich, line 73: Could you please add a description of the meaning of available and unavailable here? There is not a single word in the docs of ros2_control about that, we should also add something there: https://github.com/ros-controls/ros2_control/blob/master/controller_manager/doc/controller_chaining.rst

christophfroehlich (changes requested):
I tried add a picture of the chain with view_controller_chains, but there seems to be an issue
https://github.com/ros-controls/ros2_control/issues/1095
The example itself seem to work, so let's add the picture later once the cli script is fixed.

example_12/bringup/launch/launch_chained_controllers.launch.py
```diff
@@ -0,0 +1,51 @@
...
+                target_action=position_controller_spawner,
+                on_exit=[forward_position_controller_spawner],
+            )
+        )
+    )
+
+    nodes = [
+        position_controller_spawner,
+        delay_forward_position_controller_spawner_after_joint_state_broadcaster_spawner,
+    ]
```
christophfroehlich, line 49:
```suggestion
    delay_forward_position_controller_spawner_after_position_controller_spawner = (
        RegisterEventHandler(
            event_handler=OnProcessExit(
                target_action=position_controller_spawner,
                on_exit=[forward_position_controller_spawner],
            )
        )
    )

    nodes = [
        position_controller_spawner,
        delay_forward_position_controller_spawner_after_position_controller_spawner,
    ]
```

christophfroehlich:
Maybe one thing which is not clear for me: Why does one need the 2nd level position_controller? I tried to directly link the first-level controllers to the forward_position_controller but get
```
[controller_manager]: Can't activate controller 'forward_position_controller': Command interface with 'joint1_position_controller/joint1/position/position' does not exist
```
why?

saikishor:
> Maybe one thing which is not clear for me: Why does one need the 2nd level position_controller? I tried to directly link the first-level controllers to the forward_position_controller and it seems to work, too. Is this just for demonstration reasons I guess?

Yes exactly, this is just for demonstration purposes.

example_12/doc/userdoc.rst
saikishor: Sure, I can add that

example_12/controllers/include/passthrough_controller/passthrough_controller.hpp
saikishor: You are right. I've modified it to use the validators.

example_12/controllers/src/passthrough_controller.cpp
saikishor: Well, I used it as the convention used in the ros2_controllers repo as well, moreover, I like to separate definitions and declarations into separate files. If you don't like this, I can merge them into .cpp

example_12/doc/userdoc.rst
christophfroehlich: Yes, please add a PR for the docs of the CM :)

example_12/controllers/src/passthrough_controller.cpp
christophfroehlich: Inside this repo it is not very important because we don't release them and the header file itself won't be used by any other user.
I don't have a clear opinion against it, but which ros2_controller has this structure? Imho, all methods are defined in its' cpp source files.

christophfroehlich:
> Yes exactly, this is just for demonstration purposes.

Could you just add a single sentence about that, maybe below literal 5? Then I'm very happy to merge ;)

example_12/controllers/src/passthrough_controller.cpp
saikishor: Well, I can remove it, if needed. I don't have a problem with it.

You cna find the similar structure with, [joint_state_broadcaster](https://github.com/ros-controls/ros2_controllers/blob/master/joint_state_broadcaster/include/joint_state_broadcaster/joint_state_broadcaster.hpp), [admittance_controller](https://github.com/ros-controls/ros2_controllers/blob/master/admittance_controller/include/admittance_controller/admittance_controller.hpp) etc..

Let me know :)

saikishor:
> Could you just add a single sentence about that, maybe below literal 5?

Done :)

example_12/controllers/src/passthrough_controller.cpp
christophfroehlich: Are we talking about the same? :D What I meant is that you have defined all your functions in the header file, and the cpp file is almost empty, having only the pluginlib export macro.

saikishor: Shit! You are absolutely right. I'm under the wrong impression. I think this needs to be fixed before merging the PR. I will do it sometime today.
Sorry for all the mess. 

Thank you

saikishor: Updated!. Thanks for the valuable feedback.

christophfroehlich (approved):
Nice, thanks!

christophfroehlich:
@mergifyio backport humble

PR #331, opened 2023-07-31
Remove non-working RHEL workflow

christophfroehlich:
A follow-up to #267: Removing RHEL completely now. 

> Prolly never needed it


destogl (approved):
Thanks!

PR #339, opened 2023-08-13
Extract robot description into global package (again)

christophfroehlich:
Starting with r6bot, I don't want to copy all meshes etc. for examples sharing the same description.

For this reason, I extract all robot description files into one global package -> xacro-description macros, meshes etc. 
The xacro files including/defining the ros2_control remain in their respective example package.

- [x] rrbot
- [x] diffbot
- [x] r6bot


destogl (approved):
This looks great!

Now you also see the reason of the initial structure – as something like this is what one would usually do in classical repo.

christophfroehlich:
@mergifyio backport humble

PR #350, opened 2023-08-20
Fix diffbot position integration

emereum:
The hardware interface for diffbot (example 2) integrates motor position data using the 2nd motor's position for both motors. This causes odometry to report movement along a single axis only. This PR fixes the issue and restores expected odometry behaviour.


christophfroehlich (approved):
You are right, thanks!

christophfroehlich:
I was wondering why this never was discovered before: The diff_drive_controller uses
- the velocity feedback if `position_feedback=false`, or 
- in the default case like in this example, the change from the last position feedback devided by the time interval

for the odometry calculation, which is then transformed and published on the /tf topic.

This means, that the wrong position of the joint in the /joint_states topic isn't used anywhere.

christophfroehlich:
@mergifyio backport humble

PR #361, opened 2023-09-25
Minor fixes in documentation (example_1 and example_12)

olivier-stasse:
This PR is fixing simple stuff on top of the amazing recent work on documenting this repo. 

In example 1, the first step is optional and RViz should be closed, and the second step start also RViz. Plus there is two in as some point.

In example 12, the file README.md is giving the wrong title for the example.


christophfroehlich (approved):
LGTM, thanks! The hint to the optional first step could be included to the other examples as well.

christophfroehlich:
@mergifyio backport humble

PR #334, opened 2023-08-05
[Example 14] Modular robot with actuators not providing states

christophfroehlich:
This PR is a rebase of #147 onto master, adding it as example 14.

PR #368, opened 2023-09-29
[Example 14] Fix return type of lambda

christophfroehlich:
As commented with #334, there was an missing return type of the lambda function.


christophfroehlich:
@mergifyio backport humble

olivier-stasse (approved):
LGTM

PR #401, opened 2023-12-24
[CI] Reviewer Lottery: Use contains() in if rule

christophfroehlich:
Maybe this works now, see https://docs.github.com/en/actions/learn-github-actions/expressions#example-matching-an-array-of-strings

### DIFF
--- a/.github/workflows/reviewer_lottery.yml
+++ b/.github/workflows/reviewer_lottery.yml
@@ -6,7 +6,16 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


christophfroehlich:
@mergifyio backport iron

christophfroehlich:
I found out why it did not work earlier: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
> This event runs in the context of the **base of the pull request**, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

PR #430, opened 2024-01-15
Publish correct joint velocities in r6bot example 7

### LINKED ISSUE #427: Position data copied to velocity field of trajectory point message
Example 7's trajectory generator copies position data to the trajectory point message's velocity vector (see snippet below). The example seems to work as expected regardless so if this is intended it would be nice to have a comment explaining why this is necessary.

https://github.com/ros-controls/ros2_control_demos/blob/efe3aa2446ebc1aa43cccd586d568cb25c3b0f49/example_7/reference_generator/send_trajectory.cpp#L79-L85

Changing to copy velocity data to the trajectory point message's velocity vector doesn't change the example's behaviour so I'm assuming only the position field of the trajectory point is used by the controller.

ecmjohnson:
Resolves #427 by copying joint velocities into the trajectory point's velocity field. This makes no functional difference for the r6bot controller when both position and velocity are provided as shown by @christophfroehlich [here](https://github.com/ros-controls/ros2_control_demos/issues/427#issuecomment-1891645903).

Tested by echoing the `/r6bot_controller/joint_trajectory` topic. Prior to the change, both `velocities` and `positions` were always the same:
```
...
- positions:
  - 0.2526216532445078
  - -0.21062744421529653
  - -0.3485843399000512
  - -0.13795689568475478
  - -1.2928197138013211e-16
  - 0.25262165324450775
  velocities:
  - 0.2526216532445078
  - -0.21062744421529653
  - -0.3485843399000512
  - -0.13795689568475478
  - -1.2928197138013211e-16
  - 0.25262165324450775
  accelerations: []
  effort: []
  time_from_start:
    sec: 1
    nanosec: 924242424
...
```

After the change, the `velocities` and `positions` differ:
```
...
- positions:
  - 0.2526216532445078
  - -0.21062744421529653
  - -0.3485843399000512
  - -0.13795689568475478
  - -1.2928197138013211e-16
  - 0.25262165324450775
  velocities:
  - 0.37845782334674916
  - 0.47486669652093844
  - 0.719777222778079
  - 0.24491052625714027
  - -8.847089727481716e-17
  - 0.37845782334674904
  accelerations: []
  effort: []
  time_from_start:
    sec: 1
    nanosec: 924242424
...
```

No difference in the example is observed.

### DIFF
--- a/example_7/reference_generator/send_trajectory.cpp
+++ b/example_7/reference_generator/send_trajectory.cpp
@@ -81,7 +81,7 @@ int main(int argc, char ** argv)
       trajectory_point_msg.positions.data(), joint_positions.data.data(),
       trajectory_point_msg.positions.size() * sizeof(double));
     std::memcpy(
-      trajectory_point_msg.velocities.data(), joint_positions.data.data(),
+      trajectory_point_msg.velocities.data(), joint_velocities.data.data(),
       trajectory_point_msg.velocities.size() * sizeof(double));
 
     // integrate joint velocities


saikishor (approved):
Thank you for fixing it

saikishor:
@ecmjohnson The changes explain themself, however, out of curiosity,  did you get this message when the robot is moving? or when it is steady?. I'm asking because the velocities are non-zero

saikishor:
@ecmjohnson thanks for the information. Fine then the issue is coming from the solver itself. 
Let's wait for @christophfroehlich to review and then we can merge it

christophfroehlich:
I had a quick look. The IK solver is called with a twist in tool frame and generates a velocity vector in joint space, which is then integrated to positions. That's why the velocity is not zero in the end: one should probably add another field to the message with the last position (is discarded now) and zero velocity.
Maybe the hardware plugin should be improved in a future PR using both position+velocity fields.

christophfroehlich (commented):
Merging this one now.

saikishor:
Thank you @christophfroehlich

PR #456, opened 2024-02-29
Update all demos using the description from topic rather than the parameter

fmauch:
This will make sure things work after https://github.com/ros-controls/ros2_control/pull/1358 is merged.

I'll update this demo by demo whenever I find the time and check the demos according to their docs after modifying them.

### DIFF
--- a/example_14/hardware/rrbot_actuator_without_feedback.cpp
+++ b/example_14/hardware/rrbot_actuator_without_feedback.cpp
@@ -183,7 +183,7 @@ hardware_interface::return_type RRBotActuatorWithoutFeedback::read(
 hardware_interface::return_type ros2_control_demo_example_14::RRBotActuatorWithoutFeedback::write(
   const rclcpp::Time & /*time*/, const rclcpp::Duration & /*period*/)
 {
-  if (!std::isnan(hw_joint_command_))
+  if (std::isfinite(hw_joint_command_))
   {
     // START: This part here is for exemplary purposes - Please do not copy to your production code
     RCLCPP_INFO(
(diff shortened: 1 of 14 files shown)


example_14/hardware/rrbot_actuator_without_feedback.cpp
```diff
@@ -183,21 +183,23 @@ hardware_interface::return_type RRBotActuatorWithoutFeedback::read(
...
-    rclcpp::get_logger("RRBotActuatorWithoutFeedback"), "Writing command: %f", hw_joint_command_);
-
-  // Simulate sending commands to the hardware
-  std::ostringstream data;
-  data << hw_joint_command_;
-  RCLCPP_INFO(
-    rclcpp::get_logger("RRBotActuatorWithoutFeedback"), "Sending data command: %s",
-    data.str().c_str());
-  send(sock_, data.str().c_str(), strlen(data.str().c_str()), 0);
+  if (!std::isnan(hw_joint_command_))
```
saikishor, line 186:
```suggestion
  if (std::isfinite(hw_joint_command_))
```

May be `isfinite` is better here as it handles both `NaN` and `Infinity`

saikishor (changes requested):

fmauch:
Migrated all demos and manually tested them.

saikishor (approved):
The changes look good to me!

Let's merge once https://github.com/ros-controls/ros2_control/pull/1358 is merged!

fmauch:
> The changes look good to me!
> 
> Let's merge once [ros-controls/ros2_control#1358](https://github.com/ros-controls/ros2_control/pull/1358) is merged!

I don't see any specific reason to wait for that, as this PR removes long-deprecated mechanisms and is in line with the docs again, but sure, it's not necessarily required before https://github.com/ros-controls/ros2_control/pull/1358 is merged.

saikishor:
Sure, we can merge it. You have a point 👍🏽

christophfroehlich (approved):
LGTM, thank you for thinking of this! 
should we backport this? this should work even with humble?

fmauch:
No, this will not work on humble because we subscribe to robot_description and not ~/robot_description on rolling.

PR #481, opened 2024-03-27
Cleanup example 12

christophfroehlich:
@saikishor I just found some gazebo stuff in example 12, I think this can be cleaned up?


saikishor (approved):
@christophfroehlich Thanks for the cleanup. Yes, this can be removed. :)

PR #486, opened 2024-04-01
Update Example 9 to use Gazebo Sim and .xml launch files

destogl:
Used for testing ros-controls/ros2_control#1354.

@christophfroehlich, This should help you with #482—it should replace it if that is OK with you.

### DIFF
--- a/example_9/bringup/launch/rrbot_gazebo_sim.launch.xml
+++ b/example_9/bringup/launch/rrbot_gazebo_sim.launch.xml
@@ -1,0 +1,64 @@
+<!--
+Copyright (c) 2024, Stogl Robotics Consulting UG (haftungsbeschränkt)
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+
+Source of this file are templates in [RosTeamWorkspace](https://github.com/StoglRobotics/ros_team_workspace) repository.
+
+Author: Dr. Denis
+-->
+
+<launch>
+  <arg name="gui"
+       default="false"
+       description="Start RViz2 automatically with this launch file."/>
+
+  <!-- Load description and start controllers -->
+  <let name="robot_description_content" value="$(command '$(find-exec xacro) $(find-pkg-share ros2_control_demo_example_9)/urdf/rrbot.urdf.xacro use_gazebo_sim:=true')" />
+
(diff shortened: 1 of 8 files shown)


christophfroehlich:
@destogl except for the world file, where is the difference to #482? And why does one need the world file? At least, I updated also the description.

example_9/bringup/launch/rrbot_gazebo_sim.launch.xml
```diff
@@ -0,0 +1,64 @@
...
+    <param name="use_sim_time" value="true" />
+  </node>
+
+  <!-- Start Gazebo -->
+  <let name="gazebo_world" value="$(find-pkg-share ros2_control_demo_example_9)/worlds/empty.world" />
+  <include file="$(find-pkg-share ros_gz_sim)/launch/gz_sim.launch.py" >
+    <arg name="gz_args" value=" -r -v 4 $(var gazebo_world)"/>
+  </include>
+
+  <node pkg="ros_gz_bridge" exec="parameter_bridge" args="
```
olivier-stasse, line 41: Why not naming the node `clock_bridge` as in the python equivalent ?

example_9/bringup/launch/rrbot_gazebo_sim.launch.xml
```diff
@@ -0,0 +1,64 @@
...
+  <node pkg="controller_manager" exec="spawner" args="forward_position_controller">
+    <param name="use_sim_time" value="true" />
+  </node>
+
+  <!-- Rviz -->
+  <node if="$(var gui)" pkg="rviz2" exec="rviz2" output="log" args="-d $(find-pkg-share $(var ros2_control_demo_description))/rrbot/rviz/rrbot.rviz">
+    <param name="use_sim_time" value="true" />
+  </node>
+
+</launch>
```
olivier-stasse, line 64: Dear @destogl maybe we do not need the equivalent xml file here, as it is not in the other examples.

example_9/worlds/empty.world
```diff
@@ -0,0 +1,72 @@
...
+      name="gz::sim::systems::SceneBroadcaster">
+    </plugin>
+    <plugin
+      filename="gz-sim-contact-system"
+      name="gz::sim::systems::Contact">
+    </plugin>
+    <plugin
+      filename="gz-sim-sensors-system"
+      name="gz::sim::systems::Sensors">
+      <render_engine>ogre2</render_engine>
```
olivier-stasse, line 27: It looks like this is the only difference with the default gz-sim `empty.world` file.

Do we need it here ?

olivier-stasse (commented):
Dear @destogl and @christophfroehlich  thanks a lot for both contributions. Switching to gz-sim is definitely a plus. 

But keeping the examples simple and documented are the main target of the repo I guess.

destogl:
> @destogl except for the world file, where is the difference to #482? And why does one need the world file? At least, I updated also the description.

@christophfroehlich you are right, I just tested and we don't need world file. I agree that we can go with #482, only thing I would like to add on top of it is `.launch.xml` as they are much more simpler.

PR #502, opened 2024-04-25
Give the controller type as parameter in ctrl namespace

christophfroehlich:
This changes the location of declaring the controller type, which will be possible with https://github.com/ros-controls/ros2_control/pull/1502

I tried to write it consistently: 

- Controller types without parameters are declared in the controller_manager namespace (joint_state_broadcaster)
- Others are declared in the controllers' namespaces, which is considered to be best-practice from now on.

### DIFF
--- a/example_1/bringup/config/rrbot_controllers.yaml
+++ b/example_1/bringup/config/rrbot_controllers.yaml
@@ -5,15 +5,10 @@ controller_manager:
     joint_state_broadcaster:
       type: joint_state_broadcaster/JointStateBroadcaster
 
-    forward_position_controller:
-      type: forward_command_controller/ForwardCommandController
-
-    joint_trajectory_position_controller:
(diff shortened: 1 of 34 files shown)


saikishor:
@christophfroehlich Thanks a lot for helping :)

christophfroehlich:
Needs https://github.com/ros-controls/ros2_control/pull/1547 to work with example_15

saikishor (approved):
LGTM now!

mhubii:
> * Controller types without parameters are declared in the controller_manager namespace (joint_state_broadcaster)
> * Others are declared in the controllers' namespaces, which is considered to be best-practice from now on.

Sorry, just curious. Why is that the case?

Would it also make sense to split the parameter files into controller manager-specific configurations and controller configurations parsed by spawner? Right now, there always exists redundant information, no?

I do like the change though!

PR #503, opened 2024-04-28
[CI] Add Jazzy

christophfroehlich:
Add Jazzy to readme and add CI jobs for it. 

Jazzy+Rolling still fail because dependencies of gz_ros2_control haven't been released yet for noble.


olivier-stasse (approved):
LGTM

PR #507, opened 2024-05-01
[doc] Remove a duplicate hint

VladimirFokow:
## General

That hint repeats 2 times. 
I've removed one of them:

> * Joint names in <ros2_control> tags in the URDF must be compatible with the controller's configuration.

[here](https://control.ros.org/master/doc/ros2_control_demos/doc/index.html#quick-hints)

(It can be backported to humble as well)


christophfroehlich (approved):
Thanks!

PR #518, opened 2024-05-13
Change links to be sphinx refs

christophfroehlich:
Using sphinx references seems to be more robust. (linkcheck failed today while releasing the Jazzy version).

### DIFF
--- a/example_1/doc/userdoc.rst
+++ b/example_1/doc/userdoc.rst
@@ -453,6 +453,6 @@ Files used for this demos
 
 Controllers from this demo
 --------------------------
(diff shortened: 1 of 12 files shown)


aprotyas (approved):
This makes sense as long as the Sphinx project structure is unchanged.

christophfroehlich:
> This makes sense as long as the Sphinx project structure is unchanged.

yes, but then would the absolute paths also fail? Same for renaming the references/files.

aprotyas:
> > This makes sense as long as the Sphinx project structure is unchanged.
> 
> yes, but then would the absolute paths also fail? Same for renaming the references/files.

True, that's a challenge regardless... Anyways, LGTM!

PR #519, opened 2024-05-13
[CI] Add path-filters and fix branches for check-docs

christophfroehlich:
- Add path filters to all build jobs
- Since https://github.com/ros-controls/control.ros.org/pull/293 we need to specify `rolling` instead of `master` for the reusable workflow. 
- I also added the correct versions for the other distros now.


.github/workflows/humble-binary-build.yml
```diff
@@ -5,9 +5,29 @@ on:
...
+    paths:
+      - '**.hpp'
+      - '**.cpp'
+      - '.github/workflows/humble-binary-build.yml'
+      - '**/package.xml'
+      - '**/CMakeLists.txt'
+      - '**.xacro'
+      - '**.py'
+      - '**.yaml'
+      - 'ros2_control_demos-not-released.humble.repos'
```
saikishor, line 17: @christophfroehlich are these paths used to check for the changes in the files and then run the CI only if there are changes?

christophfroehlich: exactly. if one changes only rst files, we don't run unnecessary CI checks.

saikishor: Awesome!

saikishor: @christophfroehlich Should we add these filters in other repos as well?. In case, we touch only the documentation, so it won't launch the unnecessary CI

saikishor (approved):
The changes look reasonable to me!

.github/workflows/humble-binary-build.yml
christophfroehlich: https://github.com/ros-controls/ros2_control/pull/1521
https://github.com/ros-controls/ros2_controllers/pull/1114
;)

PR #525, opened 2024-05-26
Remove stamped parameter

christophfroehlich:
Is default true already, and will be removed by https://github.com/ros-controls/ros2_controllers/pull/1151


olivier-stasse (approved):
LGTM

PR #540, opened 2024-07-21
Add launch tests

christophfroehlich:
The old tests only test the URDF and if the view_robot.launch.py doesn't have a python syntax error and launch_ros does not complain.

The new tests for the main launch file now also check

- The python syntax is correct
- robot_state_publisher node is available
- the controller for this example is active
- joint_states topic is available and has the correct joints

closes #422 

Deactivated tests or other temporary measures: I'll push draft PRs after merging this to not forget that!

- example_14 exit codes due to #542 
- example_15 because of bug in spawners, may be solved with #502 
- [0b8382b](https://github.com/ros-controls/ros2_control_demos/pull/540/commits/0b8382bb1a5f05820bc0af4b5931bdc09a77bb4e) can be reverted once https://github.com/ros2/rclpy/issues/1320 is fixed

PR #546, opened 2024-07-29
Reactivate tests for example_15

christophfroehlich:
Follow-up to #540 

Re-evalute after #502 is merged.

### DIFF
--- a/example_15/CMakeLists.txt
+++ b/example_15/CMakeLists.txt
@@ -28,9 +28,8 @@ install(
 if(BUILD_TESTING)
   find_package(ament_cmake_pytest REQUIRED)
 
-  # TODO(christophfroehlich) deactivated because of bug in spawner
-  # ament_add_pytest_test(test_rrbot_namespace_launch test/test_rrbot_namespace_launch.py)
-  # ament_add_pytest_test(test_multi_controller_manager_launch test/test_multi_controller_manager_launch.py)
+  ament_add_pytest_test(test_rrbot_namespace_launch test/test_rrbot_namespace_launch.py)
+  ament_add_pytest_test(test_multi_controller_manager_launch test/test_multi_controller_manager_launch.py)
 
 endif()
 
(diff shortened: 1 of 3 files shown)


christophfroehlich:
@saikishor it seems that we have a bug or a race condition in the spawner:
There is no `rrbot_1.spawner_joint_state_broadcaster` launched, but two for rrbot_2 with the obvious error
`2: [spawner-10] [WARN] [1727282065.051104080] [rrbot_2.spawner_joint_state_broadcaster]: [93mController already loaded, skipping load_controller[0m`
When I uncomment the rrbot_2_launch [here](https://github.com/ros-controls/ros2_control_demos/blob/activate_test_15/example_15/bringup/launch/multi_controller_manager_example_two_rrbots.launch.py#L159), then the JSB for rrbot_1 is spawned as expected. Any ideas?

saikishor:
> @saikishor it seems that we have a bug or a race condition in the spawner: There is no `rrbot_1.spawner_joint_state_broadcaster` launched, but two for rrbot_2 with the obvious error `2: [spawner-10] [WARN] [1727282065.051104080] [rrbot_2.spawner_joint_state_broadcaster]: �[93mController already loaded, skipping load_controller�[0m` When I uncomment the rrbot_2_launch [here](https://github.com/ros-controls/ros2_control_demos/blob/activate_test_15/example_15/bringup/launch/multi_controller_manager_example_two_rrbots.launch.py#L159), then the JSB for rrbot_1 is spawned as expected. Any ideas?

Ahh I believe I might know. Both the robots use the same controller manager here?

christophfroehlich:
No, `/rrbot_1/controller_manager` vs `/rrbot_2/controller_manager`

christophfroehlich:
The problem seemed to be the `RegisterEventHandler`. We don't need it here for the spawners (or in the demos at all?). Binary build fails until next release of ros2_control.

saikishor (approved):
LGTM

PR #547, opened 2024-07-29
Revert spawner reordering

christophfroehlich:
Follow-up to #540.

Re-evalute if there is a fix for https://github.com/ros2/rclpy/issues/1320


christophfroehlich:
Now the tests passes, unsure if the linked issue got fixed or it is just seldomly flaky.

PR #548, opened 2024-07-29
Activate assert exit codes

christophfroehlich:
Follow-up to #540 

These are flaky, especially example 14 is making trouble with segfaults of the ros2_control node.


saikishor (approved):
LGTM.
Thank you! :)

PR #552, opened 2024-07-31
Remove launch testing timeout marker

christophfroehlich:
The marker `@launch_testing.ready_to_test_action_timeout(20)` was not backported to humble
https://github.com/ros2/launch/pull/625

This makes now the compatibility build fail, and also stops this from backporting see #549

### DIFF
--- a/example_10/test/test_rrbot_launch.py
+++ b/example_10/test/test_rrbot_launch.py
@@ -38,7 +38,6 @@ from launch.actions import IncludeLaunchDescription
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch_testing.actions import ReadyToTest
 
-import launch_testing
 import rclpy
 from rclpy.node import Node
 from ros2_control_demo_testing.test_utils import (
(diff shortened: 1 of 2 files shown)


example_10/test/test_rrbot_launch.py
```diff
@@ -50,11 +49,8 @@
 
 # This function specifies the processes to be run for our test
 # The ReadyToTest action waits for 15 second by default for the processes to
-# start, if processes take more time an error is thrown. We use decorator here
-# to provide timeout duration of 20 second so that processes that take longer than
-# 15 seconds can start up.
+# start, if processes take more time an error is thrown.
```
saikishor, line 52: Is this statement still true? I don't see any evaluation of checking if it takes longer or not

christophfroehlich: yes, this is part of the launch_testing
https://github.com/ros2/launch/blob/f3e9cfebcfa75c82299bb31674355a29454489fa/launch_testing/test/launch_testing/examples/ready_to_test_action_timeout_test.py#L25-L32

saikishor (approved):
LGTM

PR #553, opened 2024-07-31
Fix double quotes in python file

christophfroehlich:
The code worked on noble/rolling but didn't on iron/humble at jammy. I added this already in #549 #550 but to be save I'd also push this on the rolling version.


saikishor (approved):
LGTM

PR #417, opened 2023-12-25
[Example 13] Multi-robot system with lifecycle management 

christophfroehlich:
This is a rewrite of https://github.com/ros-controls/ros2_control_demos/pull/187 including https://github.com/StoglRobotics-forks/ros2_control_demos/pull/7.

Open questions for me:
- ~~The problems discussed in the latter PR are summarized in https://github.com/ros-controls/ros2_control/issues/1103. I'd say we can add this example even if the problem is not fixed yet.~~
- ~~It uses the rrbot URDF from example_4 and example_5. Should I generate new ones inside example 15? Currently, there is also the FTS broadcaster etc in Rviz, but maybe this wouldn't be necessary and we could just use the same Rrbot instead.~~ -> we decided in the WG meeting to reuse descriptions for the multi-robot examples.
- Why should one use a global joint state broadcaster? In the original version it was not configured properly, and all "individual" JSB published on the `/joint_states` topic and it seemed to work (better than now, without the necessity of reloading the JSB). The only difference is that with the global one, all joints are within one message. Why would someone need this for multiple robots?

PR #561, opened 2024-08-02
[Example 13] Give the controller type as parameter in ctrl namespace

christophfroehlich:
Applies changes similar to #502 to #417, but this can't be backported.


saikishor (approved):
LGTM

PR #562, opened 2024-08-03
Add guards for correct version of ros2_control

christophfroehlich:
We lately applied some breaking behavior changes of ros2_control to the demos (e.g., #502). The idea is now to check for the correct ros2_control version at compile-time

```
Starting >>> ros2_control_demo_example_1
--- stderr: ros2_control_demo_example_1                         
CMake Warning at CMakeLists.txt:20 (find_package):
  Could not find a configuration file for package "ros2_control" that is
  compatible with requested version "5.0.0".

  The following configuration files were considered but not accepted:

    /workspaces/ros2_rolling_ws/install/ros2_control/share/ros2_control/cmake/ros2_controlConfig.cmake, version: 4.13.0



CMake Error at CMakeLists.txt:23 (message):
  ros2_control version 5.0.0 or higher is required. Are you using the correct branch of the ros2_control_demos repository?
```

### DIFF
--- a/example_11/CMakeLists.txt
+++ b/example_11/CMakeLists.txt
@@ -16,6 +16,14 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   rclcpp_lifecycle
 )
 
+# Specify the required version of ros2_control
+find_package(ros2_control 4.0.0)
+# Handle the case where the required version is not found
+if(NOT ros2_control_FOUND)
+  message(FATAL_ERROR "ros2_control version 4.0.0 or higher is required. "
+  "Are you using the correct branch of the ros2_control_demos repository?")
+endif()
+
 # find dependencies
 find_package(backward_ros REQUIRED)
 find_package(ament_cmake REQUIRED)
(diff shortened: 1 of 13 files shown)


example_11/CMakeLists.txt
```diff
@@ -16,6 +16,14 @@ set(THIS_PACKAGE_INCLUDE_DEPENDS
   rclcpp_lifecycle
 )
 
+# Specify the required version of ros2_control
+find_package(controller_manager 4.0.0)
```
saikishor, line 20:
```suggestion
find_package(controller_manager 4.0.0 REQUIRED)
```

saikishor (changes requested):
May be something like this is better right?. So, you don't have to add the extra check

example_11/CMakeLists.txt
christophfroehlich: with `required` fails immediately and the error message is not printed any more. I think this might be useful for ROS beginners (which might be the major audience of the demo repositories).

christophfroehlich: if you now a way to provide a custom error message with `REQUIRED`, this would be ideal.

saikishor (approved):
LGTM

example_11/CMakeLists.txt
saikishor: You are right!. Thanks for the explanation

PR #573, opened 2024-08-22
Update publishing rate for example_2 in userdoc.rst

Pratham-Pandey:
Updated publishing rate to match with diff_drive controller update rate (as in 3rd line of [config file](https://github.com/ros-controls/ros2_control_demos/blob/master/example_2/bringup/config/diffbot_controllers.yaml)).

### DIFF
--- a/example_2/doc/userdoc.rst
+++ b/example_2/doc/userdoc.rst
@@ -88,7 +88,7 @@ Tutorial steps
 
    .. code-block:: shell
 
-    ros2 topic pub --rate 30 /cmd_vel geometry_msgs/msg/TwistStamped "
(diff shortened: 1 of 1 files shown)


saikishor (approved):

christophfroehlich:
This would not be necessary actually, the controller just uses the old command until it gets a new one :man_shrugging:

christophfroehlich:
@mergifyio backport iron humble

Pratham-Pandey:
> This would not be necessary actually, the controller just uses the old command until it gets a new one 🤷‍♂️

Would not this create unnecessary load(although negligible) on the system as publishing more frequently would use more resource?

christophfroehlich:
Sorry, forget my comment. I was wrong and thought you increased the update rate. Reducing it is totally fine, there is no reason for sending this faster here.

PR #577, opened 2024-08-23
Use new spawner option for ros args

christophfroehlich:
Changed the topic remappings to the new spawner argument proposed with https://github.com/ros-controls/ros2_control/pull/1713 

This makes it clearer instead of passing the arguments silently via the ros2_control_node.


christophfroehlich:
I added a simple test if the remapping worked.

saikishor:
> I added a simple test if the remapping worked.

That's amazing

saikishor (approved):
I like the tests. LGTM

PR #600, opened 2024-10-13
Fix the publisher parameter type to `DOUBLE_ARRAY`

christophfroehlich:
@fmauch found an issue with wrong parameter types, as we declared the type explicitly with https://github.com/ros-controls/ros2_controllers/pull/1280

```
[publisher_forward_position_controller-1] rclpy.exceptions.InvalidParameterTypeException: Trying to set parameter 'pos2' to '[0, 0]' of type 'INTEGER_ARRAY', expecting type 'DOUBLE_ARRAY': pos2
```


aprotyas (approved):

fmauch (approved):
Thanks for the quick fix. I grep-verified, that there is no `[0, 0]` in the demos repo anymore.

PR #624, opened 2024-11-03
Example3: Add section on command mode switching

christophfroehlich:
up to now this was only discussed with launch file parameters, not a command mode switch during runtime.

### DIFF
--- a/example_3/doc/userdoc.rst
+++ b/example_3/doc/userdoc.rst
@@ -142,7 +142,7 @@ Tutorial steps
       [ros2_control_node-1]   pos: 0.67, vel: 5.00, acc: 0.00 for joint 0
       [ros2_control_node-1]   pos: 0.67, vel: 5.00, acc: 0.00 for joint 1
 
-6. Now you can try to switch controllers during runtime, which also changes the command mode automatically. First, you have to load the new controller, for example the ``forward_position_controller`` if you haven't changed the launch file argument.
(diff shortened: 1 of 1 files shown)


example_3/doc/userdoc.rst
```diff
@@ -142,7 +142,30 @@ Tutorial steps
       [ros2_control_node-1]   pos: 0.67, vel: 5.00, acc: 0.00 for joint 0
       [ros2_control_node-1]   pos: 0.67, vel: 5.00, acc: 0.00 for joint 1
 
-6. To demonstrate illegal controller configuration, use one of the following launch file arguments:
+6. Now you can try to switch controllers during runtime, which also changes the command mode automatically. First, you have to load the new controller, for example the ``forward_position_controller`` if you haven't changed the launch file argument.
```
saikishor, line 145:
```suggestion
6. Now you can also switch controllers during runtime, which also changes the command mode automatically. First, you have to load the new controller, for example the ``forward_position_controller``
```

saikishor (changes requested):
A minor change

example_3/doc/userdoc.rst
christophfroehlich: forward_position_controller is already loaded if the launch file argument was set appropriately, that's what I wanted to say here?

saikishor: Ok I modified it to have the last part. What about the beginning

saikishor (approved):
LGTM

PR #654, opened 2024-11-28
Integrate pal_statistics for introspecting the ros2_control_demos

saikishor:
Needs: https://github.com/ros-controls/ros2_control/pull/1897

### DIFF
--- a/example_12/controllers/src/passthrough_controller.cpp
+++ b/example_12/controllers/src/passthrough_controller.cpp
@@ -85,6 +85,11 @@ controller_interface::CallbackReturn PassthroughController::on_configure(
   reference_interfaces_.resize(
     reference_interface_names_.size(), std::numeric_limits<double>::quiet_NaN());
 
+  for (size_t i = 0; i < reference_interface_names_.size(); i++)
+  {
+    REGISTER_DEFAULT_INTROSPECTION(reference_interface_names_[i], &reference_interfaces_[i]);
+  }
+
   return controller_interface::CallbackReturn::SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


christophfroehlich:
I'll set it to draft until a valid version is in the ros2_control repo.

saikishor:
> I'll set it to draft until a valid version is in the ros2_control repo.

Sure. Thanks 
Sorry for the noise🙏🏽🙏🏽

example_12/controllers/src/passthrough_controller.cpp
```diff
@@ -85,6 +85,11 @@ controller_interface::CallbackReturn PassthroughController::on_configure(
   reference_interfaces_.resize(
     reference_interface_names_.size(), std::numeric_limits<double>::quiet_NaN());
 
+  for (size_t i = 0; i < reference_interface_names_.size(); i++)
+  {
+    REGISTER_ROS2_CONTROL_INTROSPECTION(reference_interface_names_[i], &reference_interfaces_[i]);
```
christophfroehlich, line 90: aren't reference interfaces exported by default?

saikishor: Yes yes, they are.
I added to just show how to introspect internal variables. I didn't know what else to register in this controller

christophfroehlich: I see. I had a look now and don't find any other more useful variables to introspect in this repository. maybe we should have a look in ros2_controllers and register more variables to pal_statistics, currently most of the controllers have their own publisher with a custom status message type. not sure what is the best way in the controllers for doing so

saikishor: I would say yes, it is better to register some variables in the controllers

saikishor:
> @saikishor how do we proceed here?

I would say let's merge it at least it shows how to introspect internal variables

christophfroehlich (changes requested):
Then we need some documentation (in the code, in the rst) ;)

PR #664, opened 2024-12-15
Use class methods for rclpy.init/shutdown

christophfroehlich:
I followed the guidelines of https://docs.ros.org/en/rolling/p/launch_testing/index.html

I was not aware that `setUp`/`tearDown` will be called for every test case.


saikishor (approved):
LGTM

Thanks for the changes

Issue #640, opened 2024-11-15
Example_13 tests are flaky


christophfroehlich:
Some controllers are not spawned within time

#639

christophfroehlich:
Seems to be fixed :crossed_fingers:

PR #666, opened 2024-12-15
Use launch_utils instead of a spawner per controller

christophfroehlich:
Adressing #640, maybe this fixes the flakyness.


saikishor (approved):
Perfect.
Thanks for fixing it

PR #696, opened 2025-01-22
Update docs about realtime thread

christophfroehlich:


example_7/doc/userdoc.rst
```diff
@@ -22,7 +22,7 @@ ros2_control introduces ``state_interfaces`` and ``command_interfaces`` to abstr
...
+The main program is a realtime read, update, write loop. During the  read call, hardware drivers that conform to ``HardwareInterface`` update their offered ``state_interfaces`` with the newest values received from the hardware. During the update call, controllers calculate commands from the updated ``state_interfaces`` and writes them into its ``command_interfaces``. Finally, during to write call, the hardware drivers read values from their offer ``command_interfaces`` and send them to the hardware. The ``ros2_control_node`` runs the main loop in a realtime thread. The ``ros2_control_node`` runs a second non-realtime thread to interact with ROS publishers, subscribers, and services.
```
saikishor, line 25:
```suggestion
The main program is a realtime read, update, write loop. During the  read call, hardware drivers that conform to ``HardwareInterface`` update their offered ``state_interfaces`` with the newest values received from the hardware. During the update call, controllers calculate commands from the updated ``state_interfaces`` and writes them into its ``command_interfaces``. Finally, during to write call, the hardware drivers read values from the ``command_interfaces`` and sends them to the hardware. The ``ros2_control_node`` runs the main loop in a realtime thread. The ``ros2_control_node`` runs a second non-realtime thread to interact with ROS publishers, subscribers, and services.
```

saikishor (approved):

christophfroehlich:
@saikishor  Should we put this info in a more prominent place, instead of hiding it in the examples?

saikishor:
> @saikishor  Should we put this info in a more prominent place, instead of hiding it in the examples?

Let's document it more clearly in the control.ros.org from controller_manager docs

saikishor:
Can you add this to the roadmap?

PR #700, opened 2025-01-22
Fix remapping arg in launch file and tests

christophfroehlich:
The remap_odometry_tf launch file argument was not cleaned up properly with #577


Juliaj (approved):
LGTM.

PR #725, opened 2025-02-26
Cleanup dependencies and update metadata in package.xml

christophfroehlich:
I realized that there is a mess with the test dependencies

- I added missing `test_depend` like launch_testing, and removed unused ones.
- I removed duplicates, which were already defined in as `exec_depend`, see [the docs](https://docs.ros.org/en/rolling/Tutorials/Intermediate/Rosdep.html#test-depend).
- I reorderd some xmls to satisfy the linter


saikishor (approved):
LGTM.

Thanks for the cleanup 🙌🏽🙌🏽
As far as I can see, the dependencies look fine

PR #729, opened 2025-03-04
Print hardware component's rw_rate in example_1

christophfroehlich:
Was a simple test to see if rw_rate is set during on_init() call.


saikishor (approved):
LGTM

PR #735, opened 2025-03-09
Update index.rst

anton-matosov:
run rosdep without sudo


christophfroehlich (approved):
You are right, thanks

PR #746, opened 2025-03-17
Fix broken links and text indentation in example_16 doc

Juliaj:
Fix https://github.com/ros-controls/ros2_control_demos/issues/742 and a text indentation issue for Tutorial [steps 2](https://control.ros.org/rolling/doc/ros2_control_demos/example_16/doc/userdoc.html#tutorial-steps). 

- Changes have been tested locally by following instructions from https://github.com/ros-controls/control.ros.org?tab=readme-ov-file#build-instructions.

### DIFF
--- a/example_16/doc/userdoc.rst
+++ b/example_16/doc/userdoc.rst
@@ -8,7 +8,7 @@ DiffBot with Chained Controllers
 
 This example shows how to create chained controllers using diff_drive_controller and pid_controllers to control a differential drive robot. It extends *example_2*. If you haven't already, you can find the instructions for *example_2* in :ref:`ros2_control_demos_example_2_userdoc`. It is recommended to follow the steps given in that tutorial first before proceeding with this one.
 
-This example demonstrates controller chaining as described in :ref:`controller_chaining`. The control chain flows from the diff_drive_controller through two PID controllers to the DiffBot hardware. The diff_drive_controller converts desired robot twist into wheel velocity commands, which are then processed by the PID controllers to directly control the wheel velocities. Additionally, this example shows how to enable the feedforward mode for the PID controllers.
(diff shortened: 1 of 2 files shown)


example_16/doc/userdoc.rst
```diff
@@ -8,7 +8,7 @@ DiffBot with Chained Controllers
 
 This example shows how to create chained controllers using diff_drive_controller and pid_controllers to control a differential drive robot. It extends *example_2*. If you haven't already, you can find the instructions for *example_2* in :ref:`ros2_control_demos_example_2_userdoc`. It is recommended to follow the steps given in that tutorial first before proceeding with this one.
 
-This example demonstrates controller chaining as described in :ref:`controller_chaining`. The control chain flows from the diff_drive_controller through two PID controllers to the DiffBot hardware. The diff_drive_controller converts desired robot twist into wheel velocity commands, which are then processed by the PID controllers to directly control the wheel velocities. Additionally, this example shows how to enable the feedforward mode for the PID controllers.
```
christophfroehlich, line 11: but this ref does work, doesn't it? better to keep it, because we don't have to change this in case the section moves to a different location.

Juliaj: Interesting. The `ref` does work based on official documentation site. I've reverted the change.  The modification was initially made due to a warning about an undefined label and the link not rendering properly in my local browser environment.

Juliaj:
> Thanks for the follow up.
> 
> When you are into fixing indents, there is also a little error in the indentation here in literal 1 and 2 

Thanks for spotting these, fix made.

example_16/doc/userdoc.rst
christophfroehlich: have you cloned the ros2_control package also into the docs folder? (I always set symlinks from the control.ros.org repo to my regular workspace)

christophfroehlich (approved):
Thanks again!

example_16/doc/userdoc.rst
Juliaj: Ah, that may have caused it. I only added a symlink for the control demo git repo.

PR #751, opened 2025-03-24
Update to docker file and install description

christophfroehlich:
Use correct launch file argument in docker entry point and some more details in the installation description

image


saikishor (approved):
LGTM

PR #774, opened 2025-04-09
[Example 5] Cast to double in the force torque HW interface

saikishor:
This is needed after we merged the latest changes for boolean suuport. Floats are not casted anymore, it will only caused compilation error

https://github.com/ros-controls/ros2_control/actions/runs/14365884755/job/40278566870


urfeex (approved):
That seems fair.

christophfroehlich (approved):
Lgtm

PR #653, opened 2024-11-24
Apply automatic interface export to ex14

christophfroehlich:
Follow-up to #623, needs https://github.com/ros-controls/ros2_control/pull/1885

PR #784, opened 2025-04-16
Use joint_state_interfaces in RRBotSensorPositionFeedback

christophfroehlich:
I introduced a bug with #653: The state interfaces are joints, and not sensors -> the joints haven't been set to zero -> the integration step starts from NaN and stays there.


Juliaj (approved):
LGTM.

PR #787, opened 2025-04-22
Fix velocity limit of carlike bot

christophfroehlich:
Steering is broken since the joint_limiter is active


saikishor (approved):

saikishor:
Thank you for fixing it @christophfroehlich

Issue #793, opened 2025-04-30
Incorrect trajectory interpolation in example_7


mikelikesrobots:
**Describe the bug**
Function `interpolate_point` in `r6bot_controller.cpp` currently interpolates between `point_2` and `point_2`.

**To Reproduce**
1. Go to the file, lines 128 and 133.
2. The current maths cancels out to `point_2.positions[i]`.

**Expected behavior**
Interpolation should be between `point_1` and `point_2`. I believe the correct expression would be `delta * point_2.positions[i] + (1.0 - delta) * point_1.positions[i]`, and the equivalent for velocities.

30% between 1.0 and 2.0 would therefore be (0.3 * 2.0 + (1.0 - 0.3) * 1.0) = 1.3. This is what I would expect.

**Screenshots**
N/A
**Environment (please complete the following information):**
- `master` branch
 
**Additional context**
No extra context, just a request that I can contribute the fix if you agree with the issue.

PR #794, opened 2025-04-30
Correct interpolation of example_7 to use both points (#793)

mikelikesrobots:
Fix applied from #793 . Note that behaviour is very similar, to the point of not being noticed, due to the controller frequency reducing visible errors. This merely corrects the example to have the right maths.


christophfroehlich (approved):
That makes more sense, thanks for fixing it!

PR #804, opened 2025-05-11
Fix 6DOF arm example

sea-bass:
Thanks to debugging by @catarina-p and myself, we found issues with the 6DOF arm example.

Turns out the controller kept extrapolating forever past the final point, meaning that instead of slowing down from positive to zero velocity, the velocities kept going negative further and further.

Before this fix:
https://github.com/user-attachments/assets/af815279-818a-4922-837d-94b0208035cc

After this fix:
https://github.com/user-attachments/assets/70f4796e-38e1-4ba5-b2f2-5e402c324f2c

I also did some cleanup of the trajectory time calculations since the final point's time was not exactly 3.0 as expected.

Edit: As a side note, I am somewhat surprised that this example doesn't just use the regular joint trajectory controller that is already available in ros2_control. Although it's a nice example for people writing their own controllers.

### DIFF
--- a/example_7/bringup/launch/r6bot_controller.launch.py
+++ b/example_7/bringup/launch/r6bot_controller.launch.py
@@ -64,6 +64,7 @@ def generate_launch_description():
         executable="ros2_control_node",
         parameters=[robot_controllers],
         output="both",
+        emulate_tty=True,
     )
     robot_state_pub_node = Node(
         package="robot_state_publisher",
--- a/example_7/controller/r6bot_controller.cpp
+++ b/example_7/controller/r6bot_controller.cpp
@@ -141,10 +141,11 @@ void interpolate_trajectory_point(
   double traj_len = static_cast<double>(traj_msg.points.size());
   auto last_time = traj_msg.points.back().time_from_start;
   double total_time = last_time.sec + last_time.nanosec * 1E-9;
+  double cur_time_sec = cur_time.seconds();
 
-  size_t ind = static_cast<size_t>(cur_time.seconds() * (traj_len / total_time));
+  size_t ind = static_cast<size_t>(cur_time_sec * (traj_len / total_time));
   ind = std::min(ind, static_cast<size_t>(traj_len) - 2);
-  double delta = cur_time.seconds() - static_cast<double>(ind) * (total_time / traj_len);
+  double delta = std::min(cur_time_sec - static_cast<double>(ind) * (total_time / traj_len), 1.0);
   interpolate_point(traj_msg.points[ind], traj_msg.points[ind + 1], point_interp, delta);
 }
 
--- a/example_7/controller/r6bot_controller.cpp
+++ b/example_7/controller/r6bot_controller.cpp
@@ -145,6 +145,14 @@ void interpolate_trajectory_point(
   double cur_time_sec = cur_time.seconds();
   reached_end = (cur_time_sec >= total_time);
 
+  // If we reached the end of the trajectory, set the velocities to zero.
+  if (reached_end)
+  {
+    point_interp.positions = traj_msg.points.back().positions;
+    std::fill(point_interp.velocities.begin(), point_interp.velocities.end(), 0.0);
+    return;
+  }
+
   size_t ind =
     static_cast<size_t>(cur_time_sec * (traj_len / total_time));  // Assumes evenly spaced points.
   ind = std::min(ind, static_cast<size_t>(traj_len) - 2);
@@ -167,12 +175,11 @@ controller_interface::return_type RobotController::update(
     bool reached_end;
     interpolate_trajectory_point(*trajectory_msg_, time - start_time_, point_interp_, reached_end);
 
-    // If we have reached the end of the trajectory, reset it and set velocities to zero.
+    // If we have reached the end of the trajectory, reset it..
     if (reached_end)
     {
       RCLCPP_INFO(get_node()->get_logger(), "Trajectory execution complete.");
       trajectory_msg_.reset();
-      std::fill(point_interp_.velocities.begin(), point_interp_.velocities.end(), 0.0);
     }
 
     for (size_t i = 0; i < joint_position_command_interface_.size(); i++)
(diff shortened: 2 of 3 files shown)


example_7/controller/r6bot_controller.cpp
```diff
@@ -141,10 +141,11 @@ void interpolate_trajectory_point(
   double traj_len = static_cast<double>(traj_msg.points.size());
   auto last_time = traj_msg.points.back().time_from_start;
   double total_time = last_time.sec + last_time.nanosec * 1E-9;
+  double cur_time_sec = cur_time.seconds();
 
-  size_t ind = static_cast<size_t>(cur_time.seconds() * (traj_len / total_time));
+  size_t ind = static_cast<size_t>(cur_time_sec * (traj_len / total_time));
   ind = std::min(ind, static_cast<size_t>(traj_len) - 2);
-  double delta = cur_time.seconds() - static_cast<double>(ind) * (total_time / traj_len);
+  double delta = std::min(cur_time_sec - static_cast<double>(ind) * (total_time / traj_len), 1.0);
```
sea-bass, line 148: this is the actual fix

There are alternatives, such as just checking that `cur_time > total_time` and stopping to update the hardware interface, etc.

Open to suggestions.

mikelikesrobots: It makes more sense to me to have a special case check. When the current time is above the trajectory time maximum, set position to the last position in the trajectory and the velocity to 0.

The previous version only worked when the last two velocities and positions were 0, so the interpolation gave 0 anyway.
This new version has an issue that a trajectory ending in a non-zero velocity will continue to move at that velocity until it receives a new trajectory command

mikelikesrobots (commented):
The changes to the trajectory controller make sense, and the code to set the time I think is simpler. Nice!

mikelikesrobots:
> @mikelikesrobots did you run into this in your recent tutorial BTW? This is how we found this issue...

I didn't run into this at all! The issue I had was that it didn't always receive the trajectory, or it ignored it. When it did receive the trajectory, it did a perfect circle. Your code changes do point out an issue, though.

example_7/controller/r6bot_controller.cpp
sea-bass: Yeah, that and also maybe to reset the trajectory pointer so the subsequent update calls are effectively no-ops.

mikelikesrobots: I'd be concerned about the threading with that. I don't know if an atomic<optional<pointer>> would be overengineering it? I do prefer the idea that it's a no-op once a trajectory has been finished, though.

sea-bass:
> The issue I had was that it didn't always receive the trajectory, or it ignored it. When it did receive the trajectory, it did a perfect circle. Your code changes do point out an issue, though.

... and I didn't see that myselt. I am using Cyclone DDS instead of the default. Maybe give this a quick try on your end to see if that could be the difference?

example_7/controller/r6bot_controller.cpp
sea-bass: Done! I implemented your feedback and ran a bunch of times. At least on my end, the circle is back to perfect.

The only possible issue with threading, as you said, is if you send a new trajectory exactly at the final timestep, but that seems like something too specific to handle in a simple tutorial.

example_7/bringup/launch/r6bot_controller.launch.py
```diff
@@ -64,6 +64,7 @@ def generate_launch_description():
         executable="ros2_control_node",
         parameters=[robot_controllers],
         output="both",
+        emulate_tty=True,
```
sea-bass, line 67: Sorry, I'm outing myself as a `std::cout` debugging kind of guy :)

```suggestion
```

example_7/controller/r6bot_controller.cpp
```diff
@@ -160,14 +164,30 @@ controller_interface::return_type RobotController::update(
 
   if (trajectory_msg_ != nullptr)
   {
-    interpolate_trajectory_point(*trajectory_msg_, time - start_time_, point_interp_);
+    bool reached_end;
```
sea-bass, line 167: I assume it's not that terrible to allocate a bool in a controller loop, but I can make this a member variable if preferred.

christophfroehlich: IMHO this is fine

example_7/controller/r6bot_controller.cpp
```diff
@@ -136,15 +137,18 @@ void interpolate_point(
...
 void interpolate_trajectory_point(
   const trajectory_msgs::msg::JointTrajectory & traj_msg, const rclcpp::Duration & cur_time,
-  trajectory_msgs::msg::JointTrajectoryPoint & point_interp)
+  trajectory_msgs::msg::JointTrajectoryPoint & point_interp, bool & reached_end)
 {
   double traj_len = static_cast<double>(traj_msg.points.size());
   auto last_time = traj_msg.points.back().time_from_start;
   double total_time = last_time.sec + last_time.nanosec * 1E-9;
+  double cur_time_sec = cur_time.seconds();
+  reached_end = (cur_time_sec >= total_time);
```
mikelikesrobots, line 146: Nit: Could add an early return here

mikelikesrobots (approved):
Only 1 nit, LGTM. Ofc I'm not a maintainer so my approval doesn't actually help you!

example_7/controller/r6bot_controller.cpp
sea-bass: Done!

christophfroehlich:
> Only 1 nit, LGTM. Ofc I'm not a maintainer so my approval doesn't actually help you!

it does ;)

saikishor (approved):
The changes LGTM

Thank you for the changes @sea-bass 
Thanks for your review and follow-up @mikelikesrobots

christophfroehlich (approved):
The changes LGTM. I had problems in my local dev environment (see #805), but @saikishor confirmed that it works so let's go ahead

PR #828, opened 2025-06-06
Add timestamp to TwistStamped msg

Juliaj:
**Main changes**
When following the steps for example_11, I observed lot of warnings logged to the console due to missing timestamps in the header of the `TwistStamped` message. 

This PR updates the instructions and modifies the relevant examples to include timestamps in the `TwistStamped` message and  eliminates the warnings.

```
[ros2_control_node-1] [WARN] [1749236154.127088701] [bicycle_steering_controller]: Timestamp in header is missing, using current time as command timestamp.
[ros2_control_node-1] [WARN] [1749236154.160455387] [bicycle_steering_controller]: Timestamp in header is missing, using current time as command timestamp.
[ros2_control_node-1] [WARN] [1749236154.193923786] [bicycle_steering_controller]: Timestamp in header is missing, using current time as command timestamp.
[ros2_control_node-1] [WARN] [1749236154.227070187] [bicycle_steering_controller]: Timestamp in header is missing, using current time as command timestamp.
...
[ros2_control_node-1] [WARN] [1749236154.594182149] [bicycle_steering_controller]: Timestamp in header is missing, using current time as command timestamp.

```

### DIFF
--- a/example_16/bringup/launch/demo_test_helper.py
+++ b/example_16/bringup/launch/demo_test_helper.py
@@ -17,7 +17,6 @@ import rclpy
 from rclpy.node import Node
 from geometry_msgs.msg import TwistStamped
 from std_srvs.srv import SetBool
-from rclpy.time import Time
 
 
 class DiffbotChainedControllersTest(Node):
@@ -57,7 +56,6 @@ class DiffbotChainedControllersTest(Node):
 
     def publish_cmd_vel(self, delay=0.1):
         twist_msg = TwistStamped()
-        twist_msg.header.stamp = Time().to_msg()
         twist_msg.twist.linear.x = 0.7
         twist_msg.twist.linear.y = 0.0
         twist_msg.twist.linear.z = 0.0
@@ -66,6 +64,7 @@ class DiffbotChainedControllersTest(Node):
         twist_msg.twist.angular.z = 1.0
 
         while rclpy.ok():
+            twist_msg.header.stamp = self.get_clock().now().to_msg()
             self.get_logger().info(f"Publishing twist message to cmd_vel: {twist_msg}")
             self.publisher_.publish(twist_msg)
             time.sleep(delay)


example_16/bringup/launch/demo_test_helper.py
```diff
@@ -56,15 +56,17 @@ def set_feedforward_control(self):
...
+            twist_msg = TwistStamped()
+            twist_msg.header.stamp.sec = int(time.time())
+            twist_msg.header.stamp.nanosec = int((time.time() % 1) * 1e9)
+            twist_msg.header.frame_id = "base_link"
+            twist_msg.twist.linear.x = 0.7
+            twist_msg.twist.linear.y = 0.0
+            twist_msg.twist.linear.z = 0.0
+            twist_msg.twist.angular.x = 0.0
+            twist_msg.twist.angular.y = 0.0
+            twist_msg.twist.angular.z = 1.0
```
saikishor: Better to modify to just update the timestamp of the message in the while loop

example_11/doc/userdoc.rst
```diff
@@ -96,6 +96,9 @@ Tutorial steps
    .. code-block:: shell
 
     ros2 topic pub --rate 30 /bicycle_steering_controller/reference geometry_msgs/msg/TwistStamped "
+      header: 
+        stamp: now
```
saikishor: I didn't know that we could use `now`  like this haha
Something to learn :)

example_16/bringup/launch/demo_test_helper.py
```diff
@@ -56,15 +56,17 @@ def set_feedforward_control(self):
...
-        twist_msg.twist.linear.z = 0.0
-        twist_msg.twist.angular.x = 0.0
-        twist_msg.twist.angular.y = 0.0
-        twist_msg.twist.angular.z = 1.0
-
         while rclpy.ok():
+            twist_msg = TwistStamped()
+            twist_msg.header.stamp.sec = int(time.time())
+            twist_msg.header.stamp.nanosec = int((time.time() % 1) * 1e9)
+            twist_msg.header.frame_id = "base_link"
```
saikishor: I believe the frame_id can be left empty. We don't deal with frame of reference

saikishor (changes requested):

example_16/bringup/launch/demo_test_helper.py
Juliaj: Thanks for the feedback, updated.

example_16/bringup/launch/demo_test_helper.py
```diff
@@ -65,6 +64,9 @@ def publish_cmd_vel(self, delay=0.1):
         twist_msg.twist.angular.z = 1.0
 
         while rclpy.ok():
+            twist_msg.header.stamp.sec = int(time.time())
+            twist_msg.header.stamp.nanosec = int((time.time() % 1) * 1e9)
```
saikishor: https://docs.ros.org/en/iron/p/rclpy/rclpy.time.html#rclpy.time.Time.to_msg

@Juliaj may be here we can use the `to_msg` method?

saikishor:
Rest looks good to me

christophfroehlich (dismissed):
This is great, thanks :)

example_11/doc/userdoc.rst
```diff
@@ -96,6 +96,8 @@ Tutorial steps
    .. code-block:: shell
 
     ros2 topic pub --rate 30 /bicycle_steering_controller/reference geometry_msgs/msg/TwistStamped "
+      header:
+        stamp: now
```
christophfroehlich:
```suggestion
      header: auto
```
I wasn't aware of this, but for `std_msgs/msg/Header` this should also work (from humble on)
https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Topics/Understanding-ROS2-Topics.html#ros2-topic-pub

Juliaj: Thanks for the pointer. It does, updated to use `auto`.

example_16/bringup/launch/demo_test_helper.py
Juliaj: Thanks for the suggestion, updated.

example_16/bringup/launch/demo_test_helper.py
```diff
@@ -55,8 +56,8 @@ def set_feedforward_control(self):
         self.get_logger().info("Enabled feedforward control for both wheels.")
 
     def publish_cmd_vel(self, delay=0.1):
-
         twist_msg = TwistStamped()
+        twist_msg.header.stamp = Time().to_msg()
```
saikishor:
```suggestion
        twist_msg.header.stamp = self.get_clock().now().to_msg()
```
https://github.com/mikeferguson/ros2_cookbook/blob/main/rclpy%2Ftime.md

With `Time()` the timestamp will be zero, as there is no clock to fill in the data, so better use the Node's method to get the time : `self.get_clock().now()`

example_16/bringup/launch/demo_test_helper.py
```diff
@@ -17,6 +17,7 @@
 from rclpy.node import Node
 from geometry_msgs.msg import TwistStamped
 from std_srvs.srv import SetBool
+from rclpy.time import Time
```
saikishor:
```suggestion
```

saikishor (changes requested):

example_16/bringup/launch/demo_test_helper.py
Juliaj: You're right, the timestamp is zero with my previous change. My bad, missed this in my testing.

saikishor (approved):

PR #879, opened 2025-08-16
No gif animation with PDF

christophfroehlich:
https://github.com/ros-controls/control.ros.org/issues/487


christophfroehlich:
Does not seem to be necessary with https://github.com/ros-controls/control.ros.org/pull/616

PR #958, opened 2025-11-11
Fix ignored return value in example_12

habartakh:
When building the package ros2_control_demo_example_12, the following warning is generated: 
```
 /ros2_ws/src/ros-controls/ros2_control_demos/example_12/controllers/src/passthrough_controller.cpp:145:39: warning: ignoring return value of ‘bool hardware_interface::LoanedCommandInterface::set_value(const T&, unsigned int) [with T = double]’, declared with attribute ‘nodiscard’ [-Wunused-result]
  145 |       command_interfaces_[i].set_value(reference_interfaces_[i]);
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
```

This was fixed by explicitly ignoring the return value of the `set_value` function.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       args: ["--ignore=D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404"]
 
   - repo: https://github.com/psf/black
-    rev: 25.11.0
+    rev: 25.12.0
     hooks:
       - id: black
         args: ["--line-length=99"]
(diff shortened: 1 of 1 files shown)


bmagyar:
@habartakh could you please make that change?

habartakh:
Hello @bmagyar,

Sorry for the delay. I encountered unstable behavior with the controller manager.

For the issue, I added a `RCLCPP_WARN_EXPRESSION` log that triggers whenever a command interface fails to set a value, following patterns from the `ros2_controllers` repository.

PR #1021, opened 2026-01-21
Fix XACRO joint alignment and precision for IKFast compatibility

oguzhanbzglu:
## Description

This PR fixes the R6bot URDF to make it compatible with IKFast solver generation using tools like [pyikfast](https://github.com/cyberbotics/pyikfast). The original URDF contained joint misalignments and excessive decimal precision that prevented successful IKFast compilation.

### Problem
When attempting to generate an IKFast solver for the R6bot using [pyikfast](https://github.com/cyberbotics/pyikfast), the process fails because it cannot find solution or try to solve (waited couple hours). The main causes were:

1. **Excessive decimal precision**: Joint origin values contain too many decimal places
   - Example: `xyz="0.518777 0 0.067458"` (6 decimals)
   - IKFast/OpenRAVE does not handle large decimal numbers well, as they produce very large fractions and slow down or prevent solver generation 

2. **Joint misalignment**: Non-axis-aligned joints cause solver generation to fail
   - IKFast requires joints aligned to standard axes where possible

### Changes
- Rounded joint origin XYZ positions to 3 decimal places and replaced the joint positions
  - Before: `xyz="0.518777 0 0.067458"`
  - After: `xyz="0.519 0 0.067"`
- Evaluated RPY expression simpler
  - Before: `rpy="${pi/6+pi} 0 ${pi/3}"`
  - After: `rpy="${pi} 0 ${pi/2}"` (or `rpy="0 0 0"` where applicable)
- Realigned joints to standard axes where geometrically valid
- Kept the original kinematic chain and joint hierarchy unchanged
- No changes to robot geometry or motion behavior

**After changes mentioned above, IKFast successfully generated solution.**

This is how robot joints look:

image

### DIFF
--- a/example_7/reference_generator/send_trajectory.cpp
+++ b/example_7/reference_generator/send_trajectory.cpp
@@ -102,10 +102,27 @@ int main(int argc, char ** argv)
   auto & last_point_msg = trajectory_msg.points.back();
   std::fill(last_point_msg.velocities.begin(), last_point_msg.velocities.end(), 0.0);
 
-  pub->publish(trajectory_msg);
-  while (rclcpp::ok())
+  auto started = node->now();
+  while (pub->get_subscription_count() == 0)
   {
+    if (node->now() - started > rclcpp::Duration(10, 0))
+    {
+      RCLCPP_ERROR(
+        node->get_logger(), "No subscribers connected after waiting for 10 seconds. Exiting.");
+      return 1;
+    }
+    RCLCPP_INFO(
+      node->get_logger(), "Waiting for subscribers to connect to topic %s", pub->get_topic_name());
+    rclcpp::sleep_for(std::chrono::milliseconds(500));
   }
+
+  RCLCPP_INFO(
+    node->get_logger(), "Publishing trajectory with length %ld", trajectory_msg.points.size());
+
+  pub->publish(trajectory_msg);
+
+  rclcpp::spin(node);
+  rclcpp::shutdown();
 
   return 0;
 }
--- a/example_7/test/test_r6bot_controller_launch.py
+++ b/example_7/test/test_r6bot_controller_launch.py
@@ -30,6 +30,9 @@
 
 import os
 import pytest
+import psutil
+import subprocess
+import time
 import unittest
 
 from ament_index_python.packages import get_package_share_directory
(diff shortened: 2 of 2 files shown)


ros2_control_demo_description/r6bot/urdf/r6bot_description.urdf.xacro
```diff
@@ -77,61 +77,61 @@
...
+    <origin xyz="-0.27 0 0.0" rpy="${-pi/2} -${pi/2} ${pi/2}"/>
     <axis xyz="0 0 1"/>
     <limit effort="1000.0" lower="${-pi}" upper="${pi}" velocity="2.5"/>
   </joint>
 
   <joint name="joint_3" type="revolute">
     <parent link="link_2"/>
     <child link="link_3"/>
-    <origin xyz="0.685682 0 0.041861" rpy="0 ${pi} ${pi+pi/2}"/>
+    <origin xyz="0.686 0 -0.27" rpy="0 ${pi} ${pi+pi/2}"/>
```
christophfroehlich, line 96: you change the xacro macro in line 104, but here it is OK? I don't think this is a problem. If IKfast is not capable of importing xacro, then you have to process with xacro first and pass the plain URDF.

oguzhanbzglu: I use this xacro and take the plain urdf from the robot_description topic to parse IKfast, but the resulting URDF isn't suitable for IKFast; I'm simplifying it.

christophfroehlich: but then for sure the ${pi} macros are evaluated, and no need to change this in the xacro

oguzhanbzglu:
This is how robot is positioned in zeros after changes.
Followed the **example_7** tutorial and tested with **send_trajectory.launch.py**, everything works as expected.

image

ros2_control_demo_description/r6bot/urdf/r6bot_description.urdf.xacro
christophfroehlich: so the literal 2 in the PR description above is outdated?

oguzhanbzglu: Yes, sorry that was wrong description of the rpy expression (removed). 
Basically, IKFast could not find solution for rpy calculations (pi/6 or pi/3) thus, I made them simpler and worked.

PR #1030, opened 2026-01-31
Spawn all controllers with different options from one spawner

christophfroehlich:
Needs https://github.com/ros-controls/ros2_control/pull/2895

- splitted the yaml files into separated ones
- use a single spawner for all controllers, and use different options (parameters, remappings).


ros2_control_demos.rolling.repos
```diff
@@ -9,8 +9,8 @@ repositories:
     version: master
   ros2_control:
     type: git
-    url: https://github.com/ros-controls/ros2_control.git
-    version: master
+    url: https://github.com/pal-robotics-forks/ros2_control.git
+    version: advanced_spawner
```
christophfroehlich:
```suggestion
    url: https://github.com/ros-controls/ros2_control.git
    version: master
```

christophfroehlich: Reverting this as the upstream PR hasn't got merged yet

PR #1079, opened 2026-03-23
Use new spawner support for allow_substs

saikishor:


fmauch (approved):
Loving it :-)

example_1/bringup/config/controller_manager.yaml
destogl, line 1: can we remove this and add it directly to launch? As it is so short? And basically there are no myna parameters for CM - so probably adding from launch files might be better for simplicity.

example_1/bringup/launch/rrbot.launch.py
```diff
@@ -30,14 +31,19 @@ def generate_launch_description():
...
             # Control node
             Node(
                 package="controller_manager",
                 executable="ros2_control_node",
                 parameters=[
                     PathSubstitution(FindPackageShare("ros2_control_demo_example_1"))
                     / "config"
-                    / "rrbot_controllers.yaml"
+                    / "controller_manager.yaml"
                 ],
```
destogl, line 47: Simple as that:

```
parameters=[{"update_rate": 10}]
```

destogl (dismissed):

christophfroehlich (approved):

PR #1053, opened 2026-02-18
Add Open Duck Mini ONNX locomotion demo

Juliaj:
Addresses https://github.com/ros-controls/ros2_control/issues/2172 by adding a ros2_control demo that runs an ONNX policy for Open Duck Mini v2 locomotion in MuJoCo.   

**Features**
- ONNX-based locomotion controller (50 Hz) with velocity + head commands
- Custom hardware interface (DuckMiniMujocoSystemInterface) for foot contact detection
- state_interfaces_broadcaster for IMU, joint states, and contact sensors
- Custom message VelocityCommandWithHead

**Dependencies**
- mujoco_ros2_control
- ONNX Runtime
- Trained ONNX model (e.g. from Open_Duck_Playground)

**Run**
```
ros2 launch ros2_control_demo_example_18 example_18_mujoco.launch.py

# In another terminal:
python3 $(ros2 pkg prefix ros2_control_demo_example_18)/share/ros2_control_demo_example_18/launch/test_motions.py
```

**Questions**
We have unit tests for the motion controller and the new hardware interface. For CI to run the full demo launch, we would need to add ONNX Runtime (doable) and mujoco_ros2_control. Since mujoco_ros2_control is not yet released,  what level of launch testing is reasonable (e.g. only unit tests for now, or a separate job that installs mujoco_ros2_control from source)?  @christophfroehlich and @saikishor, your advices are appreciated. 

@bmagyar, thanks for the nudge to submit this PR.

https://github.com/user-attachments/assets/55f00d33-4a6b-4001-a9bd-d0e4b35bf290

PR #1087, opened 2026-03-27
Example 18 fixes (frequency interpolation for ONNX locomotion controller)

MDerazNasr:
Intended as a contribution toward PR #1053 (Open Duck Mini ONNX locomotion demo).

### **CMakeLists.txt**

* Changed `FATAL_ERROR` to `WARNING` + `return()` when `onnxruntime_vendor` is not found.
* The original code halted the entire workspace build if ONNX Runtime was missing, even though no other example depends on it.

### **motion_controller: linear interpolation between inference outputs**

* The controller called `run_model_inference()` every update tick. When the control loop runs faster than the training rate, this produces the zero-velocity waypoint problem (the robot stutters at each inference boundary because every waypoint starts and ends at zero velocity).
* This PR adds a frequency insertion layer. On the first active update it computes N = round(T_training / T_control). Inference runs every N ticks.
* For all other ticks, it linearly interpolates joint commands between the two most recent inference outputs.
* At 50 Hz control (matching training), $N$ equals 1 and behaviour is unchanged.
* At higher control rates the interpolation activates with no configuration needed.
* Linear interpolation is the minimal correct solution. Higher-order splines are the natural next step.

### **Dockerfile.example_18**

* New Dockerfile that builds all source dependencies (`mujoco_ros2_control`, `onnxruntime_vendor`) from the `rolling.repos` file.
* Launches the demo with `headless=true` for CI and containerized use.
* Fixes the GLFW crash that occurs when no display is available.

## Testing

* **Environment:** Built and ran in Docker (`ros:rolling-ros-base`).
* **Verification:** Controller activates, ONNX model loads (input=[1,101] output=[1,14]), and joint commands publish without errors.

### DIFF
--- a/Dockerfile/Dockerfile.example_18
+++ b/Dockerfile/Dockerfile.example_18
@@ -1,0 +1,46 @@
+ARG ROS_DISTRO=rolling
+
+FROM ros:${ROS_DISTRO}-ros-base
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+    && apt-get install -y \
+        python3-vcstool \
+        python3-colcon-common-extensions \
+        git \
+        libosmesa6 \
+        libgl1 \
+        libglu1-mesa \
+    && apt-get autoremove -y -qq \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV MUJOCO_GL=osmesa
+
+WORKDIR /home/ros2_ws/src
+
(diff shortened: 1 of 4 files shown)


christophfroehlich:
This should maybe target the original PR from @Juliaj

MDerazNasr:
Closing to reopen against Juliaj's branch as suggested. Thanks for the review.

PR #1095, opened 2026-04-04
[example_2] Add disable_commands parameter demo for MockHardware

### LINKED ISSUE #366: Add demo for disable_command of MockHardware
> I propose to close this PR as it is clearly outdated wrt to the recent work.

You are right. However, the disable command demo would be fine to include somewhere else.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_control_demos/issues/136#issuecomment-1736981564_

lakhmanisahil:
Closes #366

Adds a demonstration of the `disable_commands` parameter of `MockHardware` (`mock_components/GenericSystem`) to `example_2` (DiffBot)

---

## Background

Issue #366 was created after PR #136 became outdated due to repo restructuring.

`example_2` is the right place for this, as it already uses `mock_components/GenericSystem`. This PR simply adds `disable_commands` as an extension.

---

## What `disable_commands` does

By default, `MockHardware` mirrors every command value directly back as a state value:

```
controller sends velocity 1.0 rad/s
    → MockHardware writes state velocity = 1.0 rad/s
    → position accumulates over time
```

Setting `disable_commands: true` breaks this mirroring:

```xml
<param name="disable_commands">true</param>
```

- Commands are still accepted by the hardware plugin
- But **states never update** — they stay frozen at `0.0`
- No error is thrown — the system keeps running normally

This simulates a real hardware failure mode: commands go out successfully, but no encoder feedback comes back (e.g. broken cable, dropped fieldbus connection).

The controller manager confirms this every cycle:
```
[WARN] [controller_manager.hardware_component.system.DiffBot]: Command propagation is disabled - no values will be returned!
```

---

### `description/urdf/diffbot.urdf.xacro`
Added `disable_commands` as a top-level xacro argument (default: `false`) and passed it to the `diffbot_ros2_control` macro call. This is the entry point — xacro reads command-line arguments here and forwards them inward.

### `description/ros2_control/diffbot.ros2_control.xacro`
Added `disable_commands:=false` to the macro parameter list and embedded:
```xml
<param name="disable_commands">${disable_commands}</param>
```
inside the `MockHardware` block, alongside the existing `calculate_dynamics` param. This is where the value reaches the hardware plugin.

### `bringup/launch/diffbot.launch.py`
Added `disable_commands` as a `DeclareLaunchArgument` (default: `false`) and passed it into the `xacro Command(...)` substitution alongside `use_mock_hardware`. This exposes it on the command line.

### `doc/userdoc.rst`
Added a new paragraph directly after the `calculate_dynamics` explanation in step 6, covering:
- What `disable_commands` does conceptually
- The launch command to activate it
- Expected `ros2 topic echo /joint_states` output
- The controller manager warning that confirms it is working
- The relevant xacro configuration snippet

---

### Baseline — Normal MockHardware (positions increase)

**Terminal 1:**
```bash
ros2 launch ros2_control_demo_example_2 diffbot.launch.py use_mock_hardware:=True
```

**Terminal 2:**
```bash
ros2 topic pub --rate 10 /cmd_vel geometry_msgs/msg/TwistStamped "
  header: auto
  twist:
    linear: {x: 0.7, y: 0.0, z: 0.0}
    angular: {x: 0.0, y: 0.0, z: 1.0}"
```

**Terminal 3:**
```bash
ros2 topic echo /joint_states
```

Expected result — positions climb continuously:
```
position:
- 13.2
- 13.3   ← increasing every cycle
velocity:
- 1.0
- 1.0
```
Screenshot from 2026-04-04 07-04-39
Screenshot from 2026-04-04 07-04-42

---

### With `disable_commands` — positions frozen

**Terminal 1:**
```bash
ros2 launch ros2_control_demo_example_2 diffbot.launch.py use_mock_hardware:=True disable_commands:=True
```

**Terminal 2:** same velocity command as above

**Terminal 3:**
```bash
ros2 topic echo /joint_states
```

Expected result — positions never move:
```
position:
- 0.0
- 0.0   ← frozen forever regardless of commands
velocity:
- 0.0
- 0.0
```

**Terminal 1 (controller manager)** simultaneously shows:
```
[WARN] [controller_manager.hardware_component.system.DiffBot]: Command propagation is disabled - no values will be returned!
```

This warning repeating every 100ms confirms `disable_commands` is active and working correctly.

Screenshot from 2026-04-04 09-10-46
Screenshot from 2026-04-04 09-10-24

### DIFF
--- a/example_2/bringup/launch/diffbot.launch.py
+++ b/example_2/bringup/launch/diffbot.launch.py
@@ -39,10 +39,18 @@ def generate_launch_description():
             description="Start robot with mock hardware mirroring command to its states.",
         )
     )
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "disable_commands",
+            default_value="false",
+            description="Disable command mirroring in MockHardware, simulating a disconnected driver.",
+        )
+    )
 
     # Initialize Arguments
     gui = LaunchConfiguration("gui")
     use_mock_hardware = LaunchConfiguration("use_mock_hardware")
+    disable_commands = LaunchConfiguration("disable_commands")
 
     # Get URDF via xacro
     robot_description_content = Command(
@@ -55,6 +63,9 @@ def generate_launch_description():
             " ",
             "use_mock_hardware:=",
             use_mock_hardware,
+            " ",
+            "disable_commands:=",
+            disable_commands,
         ]
     )
     robot_description = {"robot_description": robot_description_content}
--- a/example_2/description/ros2_control/diffbot.ros2_control.xacro
+++ b/example_2/description/ros2_control/diffbot.ros2_control.xacro
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
 
-  <xacro:macro name="diffbot_ros2_control" params="name prefix use_mock_hardware">
+  <xacro:macro name="diffbot_ros2_control" params="name prefix use_mock_hardware disable_commands:=false">
 
     <ros2_control name="${name}" type="system">
       <xacro:unless value="${use_mock_hardware}">
@@ -15,6 +15,7 @@
         <hardware>
           <plugin>mock_components/GenericSystem</plugin>
           <param name="calculate_dynamics">true</param>
+          <param name="disable_commands">${disable_commands}</param>
         </hardware>
       </xacro:if>
       <joint name="${prefix}left_wheel_joint">
--- a/example_2/description/urdf/diffbot.urdf.xacro
+++ b/example_2/description/urdf/diffbot.urdf.xacro
@@ -3,6 +3,7 @@
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="diffdrive_robot">
   <xacro:arg name="prefix" default="" />
   <xacro:arg name="use_mock_hardware" default="false" />
+  <xacro:arg name="disable_commands" default="false"/>
 
   <xacro:include filename="$(find ros2_control_demo_description)/diffbot/urdf/diffbot_description.urdf.xacro" />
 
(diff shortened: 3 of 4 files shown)


saikishor (commented):
Implementation wise looks good to me.

If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?

christophfroehlich:
> Implementation wise looks good to me.
> 
> If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?

sounds useful, but then we have to patch MockHardware first

saikishor:
> > Implementation wise looks good to me.
> > 
> > If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?
> 
> sounds useful, but then we have to patch MockHardware first

Yupp!

lakhmanisahil:
> > Implementation wise looks good to me.
> > If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?
> 
> sounds useful, but then we have to patch MockHardware first

Great, thanks! What should I do next? @christophfroehlich @saikishor 

I’ll open a new issue in ros-controls/ros2_control to track adding the service to MockHardware.
Once that’s merged, I can follow up here or with a new PR demonstrating runtime toggling.
or
For now, is this PR good to merge as the static disable_commands demo?

saikishor:
> > > Implementation wise looks good to me.
> > > If we want to use it for broken communication better to use as a service right? So that way you can and turn it on and off using cli. @christophfroehlich what do you think?
> > 
> > sounds useful, but then we have to patch MockHardware first
> 
> Great, thanks! What should I do next? @christophfroehlich @saikishor 
> 
> I’ll open a new issue in ros-controls/ros2_control to track adding the service to MockHardware.
> Once that’s merged, I can follow up here or with a new PR demonstrating runtime toggling.
> or
> For now, is this PR good to merge as the static disable_commands demo?

Fine by me

saikishor (approved):

christophfroehlich (approved):
Thank you, I verified that the launch file does what described and looks great!

Issue #1118, opened 2026-05-14
Remove Path*Substition around FindExecutable


christophfroehlich:
```suggestion
                                FindExecutable(name="xacro"),
```
note for another possibility for refactoring

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_control_demos/pull/982#discussion_r3005906886_

christophfroehlich:
Fixed with #1120

PR #1120, opened 2026-05-14
Remove the extra xacro path substitution

adminlip:
Removes the unnecessary `PathJoinSubstitution` wrapper around `FindExecutable(name="xacro")` in the launch file referenced by #1118. The launch file still passes the xacro executable into the existing `Command` substitution, but no longer treats that executable lookup as a path-join operation. Verified with `python3 -m compileall -q example_7/bringup/launch/send_trajectory.launch.py` and `git diff --check`.


christophfroehlich (changes requested):
Please also fix this in example_3 and example_7

bmagyar:
As you wish my lord!

christophfroehlich (approved):
Perfect!

Issue #1119, opened 2026-05-14
Remove robot_description parameter to ros2_control_node of example_15


christophfroehlich:
```suggestion
```
this should not be necessary, robot_description parameter got deprecated and is not even used from jazzy onwards. Again, this is not releated to this PR and can be done in a different refactoring PR.

_Originally posted by @christophfroehlich in https://github.com/ros-controls/ros2_control_demos/pull/982#discussion_r3005892758_

christophfroehlich:
Closed with #1121

PR #1121, opened 2026-05-14
Remove robot_description from example 15 control nodes

mzl2233:
This removes the deprecated `robot_description` parameter from the `ros2_control_node` launches in example 15 while keeping it passed to `robot_state_publisher`. The controller manager now only receives the controller configuration, matching the follow-up requested in #1119.


bmagyar:
Thank you!

PR #1140, opened 2026-06-08
fix(example_1) : fix forward_position_controller loading on kilted

### LINKED ISSUE #1134: CI build job kilted/main failed!
An automated scheduled build failed on `master`: https://github.com/ros-controls/ros2_control_demos/actions/runs/27670020973

ShivamMaurya14:
Fixes : #1134
### Reason For Failure
In kilted, spawner no longer forwards its own node parameters to the controller_manager . Using ` parameters= ` with ParameterFile on the spawner Node means the 'type' param never reaches CM, causing fatal load failure.  I've verified from  example_9/10/12 pattern . 

After changes, local build was fine .

### Fixed

 pass config via --param-file in arguments= and hardcode joint names in YAML, consistent with example_9/10/12 pattern.

### DIFF
--- a/example_1/bringup/config/rrbot_controllers.yaml
+++ b/example_1/bringup/config/rrbot_controllers.yaml
@@ -8,6 +8,6 @@ forward_position_controller:
   ros__parameters:
     type: forward_command_controller/ForwardCommandController
     joints:
-      - $(var joint_prefix)1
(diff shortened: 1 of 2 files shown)


christophfroehlich:
this fix is not needed, as the failure comes from a missing feature in the binaries. next sync will fix that.

PR #1165, opened 2026-07-22
define 'controller_manager' as build dependency as it is used in CMake

christianrauch:
## General

Since `controller_manager` is requested in the `CMakeLists.txt`, it has to be deifned as a build dependency.

## New Examples
If you propose adding a new example, make sure that your new example has:

- [ ] The correct folder structure (described in the main [README.md](README.md))
- [ ] Example has *doc/README.rst* with description
- [ ] Detailed commands how to run an example
- [ ] Output examples of CLI commands
- [ ] Screenshots with expected visualizations (if applicable)

### DIFF
--- a/example_1/package.xml
+++ b/example_1/package.xml
@@ -22,8 +22,8 @@
   <depend>pluginlib</depend>
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
+  <depend>controller_manager</depend>
 
-  <exec_depend>controller_manager</exec_depend>
   <exec_depend>forward_command_controller</exec_depend>
   <exec_depend>joint_state_broadcaster</exec_depend>
   <exec_depend>joint_state_publisher_gui</exec_depend>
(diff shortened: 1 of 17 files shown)


christophfroehlich (approved):
You are right. We only need it for the version check there to have compile-time warnings if the wrong branch of the demos is used (people opened issues about build failures, but only compiled the wrong branch).

I see now that we haven't updated the version check for master (lyrical/rolling) yet, it still asks for version>4.0.0

saikishor (approved):

christianrauch:
Have you thought about blooming the demos? You would catch these issues at the latest on the build farm, and it is quite handy to have these demos available from the repo for simple testing.

christophfroehlich:
> Have you thought about blooming the demos? You would catch these issues at the latest on the build farm, and it is quite handy to have these demos available from the repo for simple testing.

They were not intended to be releases, but we can think of revising this decision.

PR #1183, opened 2026-08-09
fix(example_1): update include path for rrbot.ros2_control.xacro

nberrehouma:
### Description
Fixes an incorrect relative file path in `rrbot.urdf.xacro`. 

The file `rrbot.ros2_control.xacro` is located inside the `description/ros2_control/` subdirectory of the `ros2_control_demo_example_1` package rather than directly in `ros2_control/`.

### Changes
- Updated `$(find ros2_control_demo_example_1)/ros2_control/rrbot.ros2_control.xacro` to `$(find ros2_control_demo_example_1)/description/ros2_control/rrbot.ros2_control.xacro` in line 17 of `rrbot.urdf.xacro`.


christophfroehlich:
This is not correct and does not work. The files are installed in a flat hierarchy
https://github.com/ros-controls/ros2_control_demos/blob/0d9df42ab19bc3f8f53b04f8d43ae7ae9857cd64/example_1/CMakeLists.txt#L56-L59

Before you opened the PR, have you tested this at all? See our contribution guidelines..

PR #1188, opened 2026-08-26
Refactor example_1 urdf test to use urdf_test package

skunal3318:
The old test shelled out to xacro and check_urdf manually with a temp file. Swapped it for urdf_test's define_xacro_test instead, which does the same check through the xacro python library directly — no subprocess/tempfile handling needed anymore.

Also added a test matrix for the prefix arg (empty and rrbot_), since the old test only ever checked the default empty prefix and never actually exercised that param.

Tested locally — both prefix values parse cleanly and pass check_urdf. Wasn't able to run the full colcon test suite locally since this branch requires controller_manager ≥5.0.0 (Rolling) and I'm on Jazzy, but verified the actual test logic directly against the real xacro file and confirmed it matches urdf_test's define_xacro_test implementation.


skunal3318:
Looks like the binary (kilted, main) failure isn't from my change — urdf_test just isn't released for Kilted yet (E: Unable to locate package ros-kilted-urdf-test), even though it's on Jazzy and Rolling.

I could add condition="$ROS_DISTRO != 'kilted'" to the urdf_test line in package.xml so it skips on Kilted until it's released there, but I don't see that pattern used anywhere else in this repo, so didn't want to just drop it in without checking first. Let me know if that's the right approach or if you'd rather handle it a different way.

christophfroehlich:
It got released 2 days ago https://github.com/ros/rosdistro/pull/53465 Since kilted-testing is fine, this is no blocker anymore.

example_1/test/test_urdf_xacro.py
```diff
@@ -28,50 +28,23 @@
...
+from launch.actions import DeclareLaunchArgument
+from urdf_test.xacro_test import define_xacro_test
+
+xacro_file_path = Path(
+    get_package_share_directory("ros2_control_demo_example_1"),
+    "urdf",
+    "rrbot.urdf.xacro",
+)
+
+prefix_arg = DeclareLaunchArgument(
```
christophfroehlich: do we really need this prefix arg as mapping argument here?

I can imagine that we add several tests for different mappings instead

christophfroehlich (changes requested):
Thanks for picking this. 
I am not familiar with urdf_test, but from this file it seems that you only generated the description but there is no actual test here
https://github.com/pal-robotics/urdf_test/blob/humble-devel/tests/test_xacro_test.py

skunal3318:
Splitted into two explicit tests as suggested by you. Also confirmed running locally with pytest, both tests run and pass:

test_xacro_no_prefix[prefix:=] PASSED
test_xacro_custom_prefix[prefix:=rrbot_] PASSED

christophfroehlich (approved):
I simplified the test (the second input can be a map for a parameterized test, no need to create separate tests) and updated the copyright.
What do you think?

skunal3318:
Looks good, thanks. Makes sense to use a map for the parameterized version instead of two separate tests, cleaner than what I had.

PR #1198, opened 2026-09-06
Add Linux SocketCAN differential-drive hardware example

Quchaosheng:
## Summary

This PR adds Example 18, a Linux SocketCAN differential-drive hardware example for `ros2_control_demos`.

It demonstrates a Linux robot computer connecting a differential-drive base to `ros2_control` through a native non-blocking raw CAN socket, without adding a vendor-specific motor dependency to the framework.

## What is included

- `SocketCanTransport`, implemented with Linux SocketCAN (`PF_CAN`, `CAN_RAW`)
- Non-blocking socket setup and `MSG_DONTWAIT` receive/send paths
- Configurable CAN interface, defaulting to `can0` and supporting `vcan0`
- Differential-drive wheel velocity command and position/velocity state interfaces
- Replaceable `CanTransport` boundary for tests without a CAN adapter or motors
- URDF/Xacro, controller configuration, launch file, README, and user documentation
- C++ transport tests and a Python URDF/Xacro test

## Protocol used by this generic example

The protocol is intentionally simple and documented so that it can be replaced by a real motor-controller protocol:

- Command frame: `0x201`
- Feedback frame: `0x181`
- Bytes `0..3`: signed little-endian `int32`, left wheel velocity in milliradians/sec
- Bytes `4..7`: signed little-endian `int32`, right wheel velocity in milliradians/sec

## Failure and boundary semantics

- `EAGAIN`/`EWOULDBLOCK`: no new feedback; retain the last valid state.
- No valid feedback for `feedback_timeout_sec` (0.5 seconds by default): return `ERROR` instead of integrating stale velocity forever.
- Short frames, unexpected CAN IDs, and CAN error frames: reject the frame; CAN error frames return `ERROR`.
- Socket errors, a closed socket, or failed non-blocking sends: return `ERROR`.
- Non-finite command or feedback values: reject in the hardware plugin.
- Negative/non-finite control periods: return `ERROR`.
- Command values are rounded to milliradians/sec and saturated to the signed 32-bit wire range. Non-finite values encode as zero in the low-level helper, while the hardware plugin rejects them before sending.
- `on_deactivate()` sends a best-effort zero command and always closes the socket.

## Real-time considerations

- Socket operations do not block the controller loop.
- No string construction or dynamic allocation is performed in the steady-state `read()`/`write()` paths; interface names are prepared during initialization.
- CAN bitrate and link state are deliberately not configured by the node. The operator must configure `can0`/`vcan0` before launch.
- A production driver should add a device-specific watchdog/heartbeat policy and confirm the motor controller's own command timeout.

## Testing

- Encode/decode signed milliradian round trip
- Saturation and non-finite wire-value behavior
- Fake transport test without CAN hardware
- URDF/Xacro validation
- Linux `g++ -std=c++17 -Wall -Wextra -Werror` syntax check

## How to try it

```bash
sudo modprobe vcan
sudo ip link add dev vcan0 type vcan
sudo ip link set vcan0 up
ros2 launch ros2_control_demo_example_18 diffbot.launch.xml can_interface:=vcan0
```

## Scope

This is a framework demo, not a vendor driver. Real deployments should replace the example IDs/payload, add protocol-level validation and watchdogs, and test against the target motor controller.


Quchaosheng:
Follow-up hardening pushed in commit 0dab8a7: the package now explicitly skips non-Linux builds, exposes a protected transport factory for injection, guards null/closed transport use, tests send-before-open and saturation/non-finite wire values, validates finite periods/feedback timeout, and documents that bus-off is detected and propagated rather than automatically recovered.

Quchaosheng:
Design discussion for this implementation is now tracked in #1199. This PR is the concrete implementation; the Issue covers acceptance of the new demo, Linux-only boundary, transport abstraction, and failure semantics.

Quchaosheng:
Additional hardening pushed in commit e0dce54: repository-style formatting and copyright headers, explicit Linux-only build guard, transport/plugin lifecycle guards, fake-transport injection seam, saturation/non-finite/send-before-open tests, and deactivation zero-command failure warning.

Quchaosheng:
@bmagyar @destogl @moriarty @Maverobot @arne48 — bumping this one, it has been up for two weeks.

Heads up on the CI first: all five workflows (Build Rolling Dockerfile, Pre-Commit - Rolling, Rolling Binary Build, Rolling Check Docs, Rolling Semi-Binary Build) are still sitting at "This workflow is awaiting approval from a maintainer", so there is no signal on this PR at all. Approving those would help more than anything I can write here.

Second, and sorry about this one: there is a real build error in it that only shows up once the install step runs. `example_18/CMakeLists.txt` installed

  description/launch description/ros2_control description/urdf

but this package has no `description/launch` — its launch files are in `bringup/launch`. I had copied that line from example_16 / example_17, and both of those do have `description/launch`. CMake aborts with

  file INSTALL cannot find ".../description/launch": No such file or directory

which kills the install step and takes the binary job with it. Fixed in 4e5abec (one word off that line), verified with a from-scratch build of the package: it installs `urdf/`, `ros2_control/` and `launch/` cleanly now, the `test_socketcan_transport` gtest is 4/4, `xacro` + `check_urdf` on the installed `diffbot.urdf.xacro` is fine, and `pre-commit` is green (that commit also picks up the copyright headers, header guards and a using-directive that cpplint wanted).

The design discussion for the example itself — Linux-only boundary, the transport abstraction, failure semantics — stays in #1199 if anyone would rather have that conversation there.

